PR #291, opened 2026-04-08
fix: highlight selected version in changelog dialog

mvanleijenhorst:
## Summary

- Fixes the changelog version list not highlighting the selected version when clicking between versions
- Adds an `onclick` handler to toggle the `active` class and `aria-current` attribute client-side, since HTMX only swaps the content pane without re-rendering the version list

Closes #287

## Test plan

- [ ] Open the changelog dialog
- [ ] Click on different versions — the clicked version should be visually highlighted (bold + muted background)
- [ ] On initial open, the latest version should be highlighted


whit3st:
This works, but I wonder if its better/cleaner fix to render the entire dialog body as a Thymeleaf fragment and swap it on version changes, instead of only replacing the changelog content?

whit3st (approved):

PR #293, opened 2026-04-09
feat: add $formatNumber function with point and comma notation

mvanleijenhorst:
## Summary

- Add `$formatNumber(value, pattern)` JSONata custom function for formatting numbers in templates
- Support **decimal point** notation (`#,##0.00` → `1,234.56`) and **decimal comma** notation (`#.##0,00` → `1.234,56`)
- Auto-detect notation from the pattern — the last separator character determines the decimal separator
- Expression dialog shows grouped format presets (Decimal point / Decimal comma) for number and integer fields
- Builder mode supports number formatting alongside existing date formatting

## Test plan

- [ ] Verify `pnpm --filter @epistola/editor test` passes (1078 tests)
- [ ] Verify `./gradlew :modules:generation:test --tests JsonataEvaluatorTest` passes
- [ ] Open template editor, click expression chip for a number field — format dropdown should appear with point/comma presets
- [ ] Select a comma notation format (e.g., `#.##0,00`) — preview should show European formatting
- [ ] Generate PDF with `$formatNumber(subtotal, '#.##0,00')` — verify comma notation in output

Closes #289


apps/epistola/src/main/resources/templates/templates/editor.html
```diff
@@ -118,6 +119,7 @@
             template: window.TEMPLATE_MODEL,
             dataExamples: window.DATA_EXAMPLES,
             dataModel: window.DATA_MODEL,
+            variantLanguage: window.VARIANT_ATTRIBUTES?.language,
```
sdegroot, line 122: I understand why you did this. I think this introduces some challenges though:

1. An implicit attribute called language - in a system where anybody can define any language
2. The language attribute can contain literally any value, which could cause issues. This attribute should be limited the potential values which corresponds to locales
3. What about American English vs British English? Its not very relevant at this moment, but I'd prefer to design for this ahead of time (or it will be painful later)

sdegroot: I guess the attribute direction is not a bad one, because generally people will want to select on these values. However, it would require to have a certain set of attributes that are system defined (but optional to use).

Also, the editor should probably not be aware of attributes or variants. The editor itself is agnostic of the notion of variants and attributes.

mvanleijenhorst:
Is delayed and needs system attribute "Language" that is not yet developed

PR #338, opened 2026-04-26
fix: margins not rendering in PDF for px and sp values

mvanleijenhorst:
Refs #323 (specifically: "margin-top op text veld wordt niet in de pdf preview toegepast")

## Summary

- **Backend dropped px margins**: `StyleApplicator.parseSize` only recognized `pt`, `mm`, and `sp`. Templates with legacy `px` margin/padding (left over from before the `sp`/`pt` unit migration) fell through to `toFloatOrNull()` on the whole string and silently returned null, so `setMargin*()` was never called. Now parses `px` at 96 DPI (`1px = 0.75pt`).
- **Editor canvas ignored sp margins**: The canvas wrote resolved style values straight into CSS via `styleMap`. Browsers don't understand `sp`, so values like `marginTop: 2sp` rendered nothing in the editor — making editor and PDF disagree. The new `toStyleMap` helper rewrites every `Nsp` token to absolute `pt`, using the theme's spacing unit.
- **Inspector kept legacy px alive**: When a node's existing margin used a unit no longer offered by the inspector (e.g. `px` after the `sp`/`pt` switch), the dropdown showed an allowed option but the spacing input internally still used the old unit — every new edit re-saved as `Npx`. `renderSpacingInput` now migrates to the first allowed unit on read, converting displayed values accordingly. The first user edit then transparently migrates the whole spacing block.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Refactoring (no functional changes)
- [ ] Chore (dependencies, CI, tooling)

> Refactoring: extracted `toStyleMap`/`camelToKebab` from `EpistolaCanvas.ts` into a testable `style-css.ts` module; reshaped `convertSideValue` around a shared `toPt` helper so additional unit conversions stay localized.

## Component(s) Affected

- [x] Backend (Spring Boot/Kotlin)
- [x] Frontend (Editor/TypeScript)
- [ ] Documentation
- [ ] CI/CD

## Checklist

- [x] My code follows the project's code style (ktlint, EditorConfig)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix is effective or my feature works
- [x] New and existing tests pass locally (gradle test)
- [ ] I have updated the documentation if needed
- [x] I have updated the CHANGELOG.md if this is a notable change
- [x] My commits follow Conventional Commits

[...]

PR #346, opened 2026-04-28
fix: pdf rendering issues (#323)

whit3st:
## Description

Fixes four unchecked items from #323 discovered during template creation:

1. **Template deletion discoverability** — The "Delete Template" action was buried under Settings > Danger Zone. Added the delete button to the `.page-header` on all template detail tabs (Variants, Deployments, Data Contract, Settings). Extracted the form into a reusable `fragments/template-actions.html` fragment. Added an `aria-label` with the template name. Fixed a CSS bug where `button[type='submit']` globally overrode `.btn-destructive`; now scoped to `:not([class*='btn-'])`.
2. **Negative sizing enforcement** — Users could enter negative values for width, height, font size, spacing, etc., which broke layout or caused iText errors. Added `min="0"` and `Math.max(0, ...)` clamping in the editor's `renderUnitInput` and `renderSpacingInput`. The `parseValueWithUnit` regex no longer accepts a leading minus sign. On the backend, `StyleApplicator.parseSize` and `ImageNodeRenderer.parseToPt` now clamp via `coerceAtLeast(0f)`.
3. **Expression editor `{{` hint** — Users didn't know they could type `{{` directly in the rich text editor to insert expressions inline. Added a subtle hint to the text block header: "type `{{` for expressions".
4. **Header/footer borders not rendering** — `PageHeaderEventHandler` and `PageFooterEventHandler` rendered slot children directly on a `Canvas`, ignoring the node-level styles (borders, background, padding). Both handlers now wrap slot children in a `Div` and apply styles via `StyleApplicator.applyStylesWithPreset`, so footer/header borders and backgrounds render correctly in the PDF.

## Related Issue(s)

Relates to #323

## Type of Change

## Component(s) Affected

## Checklist

## Additional Notes

- Tests added: `TemplateDeleteButtonUiTest`, `StyleApplicatorTest` (negative size clamping). Existing `PageHeaderFooterTest` continues to pass.
- The `{{` hint and header/footer border fixes are new additions not yet reflected in `CHANGELOG.md` — happy to add them if needed before merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - **Contract publish impact preview**: Breaking contract publishes show a confirmation dialog listing breaking changes and affected template versions with their active environment deployments.
 - **Deployment matrix error handling**: Contract compatibility errors shown inline in the deployment matrix instead of generic "An error occurred" message.
 - **Contract usage overview**: Dialog showing all template versions with their contract version (color-coded: green=current, amber=outdated) and active deployments.
+
 ### Fixed
 
 - **Template deletion discoverability**: The "Delete Template" button now appears in the page header on all template detail tabs, not just buried in Settings > Danger Zone. Extracted the delete form into a reusable fragment (`fragments/template-actions.html`). Added `aria-label` with the template name for accessibility. Fixed a CSS bug where `button[type='submit']` globally overrode `.btn-destructive` styles — now scoped to buttons without an explicit `.btn-*` class.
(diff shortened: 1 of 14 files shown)


sdegroot:
I think regarding the borders: @mvanleijenhorst  also has some fixes for that in his PR

whit3st:
> I think regarding the borders: @mvanleijenhorst also has some fixes for that in his PR

Checked the PR #338 and what i see is margin not borders. I specifically skipped margin problems along with text align issue because they are related. I can wait until #338 is merged and adjust the PR if that's okay.

mvanleijenhorst:
Maybe a good practice to split this PR in context seperate parts

- Template deletion discoverability
- Negative sizing enforcement 
- Expression editor {{ hint 
- Header/footer borders not rendering 

The advantage is first the PR are smaller and easier to read. 

Second if the is an issue, you can easy look what commit has to do with it. 

Third if something is done in a commit, you have single context, instead of multiple and guessing if it related.

sdegroot:
I should also take that advice haha

whit3st:
Yeap, completely agree. I will do that once i am home. 

In this case, is best approach here to create individual branches and cherry pick commit(s) or is there a better way?

whit3st:
Divided the PR into #348 #349 #350 and #351

mvanleijenhorst:
> 



> Yeap, completely agree. I will do that once i am home.
> 
> In this case, is best approach here to create individual branches and cherry pick commit(s) or is there a better way?

Yes I think that is the best way.

PR #375, opened 2026-05-06
feat(stencils)!: placeholders with two-slot default/fill semantics

sdegroot:
## Summary

- Adds the `placeholder` node type to stencils — templates can override parts of a stencil's layout while the rest stays frozen.
- **Two slots per placeholder**: `default` (stencil author's content) and `fill` (template override). The PDF renderer falls back to `default` when `fill` is empty, so clearing an override reveals the default. Editor canvas dispatches on context: stencil-author/draft mode shows `default` for editing; template-fill mode shows `fill` (with the default rendered greyed-out + non-interactive as a preview when fill is empty).
- **Recursion impossible**: enforced in the stencil picker, DnD `canDropHere`, server validators (5 rules wired into 6 commands), and a defensive ancestor stack in the PDF renderer.
- **Stencil version upgrades preserve template overrides by placeholder name** (server `StencilContentReplacer.upgradeStencilInstances` + client `preserve-fills.ts` mirror).
- **Edit guard**: when a newer published version is available, "Start Editing" is hidden in the inspector with a hint to upgrade first — prevents the broken-state bug where editing stale local content rolls back a newer version on save.
- **Start Editing** now fetches the actual stencil draft from the backend and swaps the local view, capturing template overrides first and splicing them back on Publish or Discard. The Upgrade button uses the same capture/restore so the client matches the server's preservation behaviour.
- **Locked stencil layouts now actually reject drops.** Previously `pointer-events: none` was the only gate; the new `isInLockedStencilLayout` check is authoritative.
- **Auto-named placeholders**: dropping a placeholder fills `placeholder-N` (next free) so a fresh placeholder saves without manual editing.

## Breaking changes

- `UpdateStencilInTemplate` now returns `UpdateStencilInTemplateResult(upgradedCount, droppedFills)` instead of `Int?`. The HTTP handler's JSON shape becomes `{ upgraded, droppedFills: { ...by-instance-id... } }`. Pre-prod — no migration provided.
- Placeholder document model changes from one slot (`fill`) to two (`default`, `fill`). Pre-prod — no migration; any documents authored with the old single-slot shape would need re-creating.

## Docs

- New: [`docs/adr/0001-stencil-placeholders.md`](docs/adr/0001-stencil-placeholders.md) — design space (A1/A2/B/C/D), decision, addendum noting the post-v1 refinement to two slots.
- New: [`docs/stencil-placeholders.md`](docs/stencil-placeholders.md) — the chosen design (document model, editor behaviour, recursion layers, upgrade path, forward-compat for stencil parameters).
- `CLAUDE.md` — new house rule: every editor component registration must include at least one entry in `examples[]`.

## Test plan

- [x] `pnpm --filter @epistola/editor test` — **1399 vitest tests pass** (12 new: ancestry slot-aware lock + `placeholderContext`; preserve-fills round-trip; extractSubtree fill stripping; canDropHere lock-state cases; canDropHere placeholder/stencil rules).
- [x] `pnpm --filter @epistola/editor build` — clean (registry includes the new `placeholder` descriptor).
- [x] `./gradlew :modules:epistola-core:unitTest` — green; PlaceholderValidator (16 tests), StencilContentReplacerTest (11 tests including 2 new for two-slot semantics).
- [x] `./gradlew :modules:epistola-core:integrationTest` — green; `StencilPlaceholderIntegrationTest` covers 7 end-to-end flows including "template override survives a stencil-edit-and-publish cycle".
- [x] `./gradlew :modules:generation:test` — green; `StencilPlaceholderRendererTest` rewritten for two-slot model (3 tests: fill renders, default fall-through, both-empty).
- [x] `./gradlew ktlintCheck` — clean.
- [x] `pnpm format:check` — clean.
- [ ] Manual UI verification — open the editor, create a stencil with placeholder + default text, embed in a template, override the fill, edit the stencil, publish, confirm overrides preserved across the cycle. Not run by me; user to verify.


whit3st:
Clicking placeholder to focus is working in a weird way. I think previously, if stencil is locked, I wasn't able to select placeholder. Currently I can but weirdly, only bottom-half of the placeholder accepts this click. element with this selector : ```canvas-placeholder-fill canvas-placeholder-fill--empty``` accepts clicks only. Clicking any other place on the placeholder focuses the parent stencil instead.

sdegroot:
fixed that as well, thanks for the bugreport

PR #379, opened 2026-05-08
feat(forms): auto-generate slug from name on create forms

### LINKED ISSUE #376: [Feature]: when i type the name/title of a resource, I expect the slug to auto update with a similar slug variant of the input
### Problem / Motivation

Right now I always have to fill in a name/title and a slug manually. This is more effort than necessary.

### Proposed Solution

Always put the title first, slug second.

- When typing in a title, a slug version of that title should appear while i am typing
- Whenever i manually change the slug, it should no longer update the slug if i change the title 

### Alternatives Considered

_No response_

### Component

Frontend (Editor/TypeScript)

### Additional Context

_No response_

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't been requested
- [x] I have read the [Contributing Guidelines](../../CONTRIBUTING.md)

whit3st:
## Description

Adds client-side slug auto-generation to all create forms. When the user types a name/title, the slug field auto-fills with a slugified version (using the same `nameToSlug` logic already present in the stencil picker dialog). If the user manually edits the slug, further name changes stop updating it.

Implements a reusable `slug-auto.js` (IIFE) that wires up any input with `data-slug-source` pointing to a slug input. The script is included via the footer fragment so it's available on every page.

## Related Issue(s)

Closes #376

## Type of Change

## Component(s) Affected

## Checklist

- [ ] I have updated the CHANGELOG.md if this is a notable change

## Additional Notes

### DIFF
--- a/apps/epistola/src/main/resources/static/js/slug-auto.js
+++ b/apps/epistola/src/main/resources/static/js/slug-auto.js
@@ -1,0 +1,70 @@
+(function () {
+  var ALLOWED_KEYS = ['Backspace', 'Delete', 'Enter', 'Escape', 'Tab', 'Home', 'End'];
+
+  function nameToSlug(name) {
+    return name
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, '-')
+      .replace(/^-|-$/g, '');
+  }
+
+  function isValidSlugKey(key) {
+    return key.length === 1 && /^[a-z0-9-]$/.test(key);
+  }
+
+  function wireup(nameInput) {
+    var slugId = nameInput.getAttribute('data-slug-source');
+    var slugInput = document.getElementById(slugId);
+    if (!slugInput) return;
+
+    var slugManuallyEdited = false;
+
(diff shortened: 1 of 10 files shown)


sdegroot:
lets also ensure that users cannot add a slug that is invalid (regex). Maybe even upon keydown that it simply does not allow capital letters / etc.

that may be a different PR or you can put in this one.

sdegroot:
I worry that this implementation will not work well with HTMX. Can you confirm that it will apply those rules even if a form is dynamically loaded?

sdegroot:
nice, i think you missed the add variant option - not sure if there are any other places

whit3st:
I shouldn't have git reset then make it one commit. I don't know why I did that. Regarding add variant option, it is different then these forms. Slug is required and display name is optional. Currently touched forms, both are required so we can derive slug from display name. If you wanna change that for add variant form, I can either do that in another branch and open a PR or do it here, match the add variant form with other forms that creates new things. Let me know

Issue #391, opened 2026-05-11
Harden CodeListClient URL fetch: enforce Accept header + max response size


sdegroot:
The code lists feature ships a `CodeListClient` that fetches entries from an
HTTPS source URL on `RefreshCodeList`. The fetch is currently underspecified
— there is no `Accept` header, no response size cap, and no content-type
validation. A malicious or misconfigured source could return arbitrary
content, an enormous payload, or an HTML error page that Jackson then
refuses to parse with a confusing error.

## Current behaviour

In `modules/epistola-core/src/main/kotlin/app/epistola/suite/attributes/codelists/service/CodeListClient.kt`:

```kotlin
private fun fetchHttp(url: String, authType: AuthType, credential: String?): List<CodeListEntry> = catalogRestClient.get()
    .uri(url)
    .applyAuth(authType, credential)
    .retrieve()
    .body(SPRING_ENTRIES_TYPE)
```

No `Accept`, no length check, no content-type check.

## Changes wanted

1. **Send `Accept: application/json`** so cooperating servers can negotiate
   format and so we fail fast on servers that won't speak JSON.
2. **Reject responses whose `Content-Type` doesn't start with
   `application/json`** (also accept `+json` suffixes per RFC 6839).
3. **Cap the response body to 50 KB.** Read into a length-bounded buffer
   (e.g. `InputStream.readNBytes(50 * 1024 + 1)`) and reject if more bytes
   are available. Also honour `Content-Length` early-rejection when the
   server provides it.
4. **Surface clear errors via `CodeListFetchException`** for each of the
   above so the value lands in `code_lists.last_refresh_error` and the UI
   detail page shows it.
5. **Make the limit configurable** via `epistola.codelists.max-fetch-size`
   with a 50 KB default — same shape as the existing
   `epistola.catalog.max-zip-size` property.

## Tests

Add `CodeListClientSecurityTest` against a stub HTTP server (Spring's
`MockRestServiceServer` or WireMock — whichever fits the existing
`catalogRestClient` test setup):

- accepts a small `application/json` body,
- rejects a `text/html` body,
- rejects a body that exceeds 50 KB (both via `Content-Length` and via
  actual byte count for chunked responses),
- the rejection messages all surface through `CodeListFetchException` and
  land in `last_refresh_error` end-to-end (covered by an integration test
  using `RefreshCodeList`).

## Out of scope

- Authentication hardening (already supported via `AuthType`).
[...]

PR #392, opened 2026-05-11
feat(attributes): code lists as a third value constraint

### LINKED ISSUE #394: Code lists: close test gaps (UI Playwright, RefreshCodeList, subscribed-catalog refusal)
Test coverage for the code lists feature (#392) is solid on the backend domain layer (28 integration tests across `CodeListCommandsTest`, `CodeListClientTest`, `AttributeCodeListBindingTest`, `AttributeValidationCodeListTest`) but has three real gaps that should be closed before the feature is considered production-ready.

## 1. `CodeListUiTest` (Playwright) — biggest gap

Nothing in the UI is exercised end-to-end. The radio-button styling bug fixed in `f9833c5c` would have been caught by even a smoke test; everything else (the three-way constraint picker JS toggle, the HTMX hide-entry toggle, the entries-table inline editor, the catalog filter, the variant dropdown options) is untested.

Suggested coverage:

- Create an INLINE code list with a couple of entries via the new form.
- Bind a new attribute to it via the three-way picker on `/attributes/new`.
- Open a template detail page, create a variant, confirm the attribute's dropdown renders `code — label` from the bound list.
- Submit a value that's not in the list — confirm validation error surfaces with the list's display name.
- On the code-list detail page, hide an entry — confirm the row's "Hidden" badge appears and the entry disappears from the variant dropdown on reload.
- Attempt to delete the list while the attribute is bound — confirm the friendly `CodeListInUseException` message renders.

Lives under `apps/epistola/src/test/.../ui/CodeListUiTest.kt`, extends `BasePlaywrightTest` (same convention as `VariantCardUiTest`).

## 2. `RefreshCodeListTest` (backend, with a stub HTTP server)

`RefreshCodeList` is wired through but never exercised end-to-end. The classpath fetch is covered by `CodeListClientTest`, but the HTTPS path — auth headers (`BEARER`, `API_KEY`), JSON parsing of the wire format, atomic entry replacement, error recording on fetch failure — isn't.

Suggested coverage using `WireMock` or `MockRestServiceServer` (whichever fits the existing `catalogRestClient` setup):

- Happy path: configure a URL code list, refresh, confirm entries land in `code_list_entries` and `last_refreshed_at` advances.
- Failure path: stub a 500 / connection refusal — confirm existing entries are **kept**, `last_refresh_error` is populated, and the command returns the updated `CodeList` (no exception out of the command).
- Auth headers: stub a server requiring `Authorization: Bearer ...` and confirm `RefreshCodeList` sends it when the list's `auth_type = BEARER`. Same for `API_KEY` and `X-API-Key`.
- `INLINE` refusal: refreshing an INLINE list throws `CodeListNotRefreshableException`.

## 3. `SubscribedCatalogReadOnlyTest` for code lists

The CRUD commands call the existing `requireCatalogEditable` helper, which is tested for other entity types but not specifically for code lists. Each of `Create/Update/Delete/Refresh/UpdateCodeListEntryHidden` should reject on a SUBSCRIBED catalog. Small but cheap — one parameterized test covers all five.

## Out of scope

- Tests already covered by #391 (URL hardening — `Accept` header, size cap, content-type validation). #391 will likely add a `CodeListClientSecurityTest`; that file is naturally adjacent to the `RefreshCodeListTest` proposed here but should stay focused on the hardening rules.
- Tests for the variants UI rework (#393) — those land with that PR.
- Concurrent-refresh tests (low priority; the transaction is atomic per reader).

sdegroot:
## Summary

- New **code lists** entity: tenant + catalog scoped collections of `{code, label, hidden}` entries an attribute can bind to as an alternative to free-format and inline `allowedValues`. Sourced from a curated inline editor or fetched over HTTPS with manual refresh; per-entry `hidden` flag soft-deprecates entries (still accepted by validation, filtered from pickers).
- Variant attribute form (create + edit dialogs) and filter bar now render dropdown options from code list entries when bound, with `code — label` display when the two differ.
- New UI at `/tenants/{id}/code-lists/**` (list, create, detail with hide-entry toggle and refresh-from-source). Three-way "Value constraint" picker replaces the old single textarea on the attribute form. Nav gains a "Code lists" entry.

## Schema (V29)

```
code_lists                              (tenant_key, catalog_key, slug)
code_list_entries                       FK ON DELETE CASCADE to code_lists
variant_attribute_definitions.code_list_*  composite FK ON DELETE RESTRICT,
                                           XOR check vs allowed_values
```

(Renumbered from V28 after merging main, which took V28 for the stencil parameters migration.)

Validation hits a single SQL existence check on the variant-write hot path. Cross-catalog binding within a tenant is allowed (shared `tenant_key` enforces same-tenant); cross-tenant is impossible.

## Test coverage

- 🟢 **Backend domain** — 38 integration tests covering CRUD, validation, bindings, cross-catalog binding, hidden-entry semantics, FK enforcement, SUBSCRIBED-catalog refusal, URL-fetch happy path + failure path + auth headers + INLINE refusal.
- 🟢 **UI** — 8 Playwright tests covering the inline-entries new form, the three-way constraint picker, variant dropdown rendering with `code — label`, hidden-entry filtering in the dropdown, and the list-page catalog filter.
- 🟡 **Schema migration** — exercised by every integration test running against a fresh DB; no direct assertion of the XOR check or partial index.

Closes #394 (test gaps).

## Out of scope / deferred

Documented in CHANGELOG and `docs/code-lists.md`. Need a coordinated change in `epistola-contract` so deferred to a follow-up:

- Catalog-qualified variant attribute references (today `validateAttributes` looks up by slug only — silent collision when the same slug exists in multiple catalogs of a tenant).
- Bundled "system" catalog with reserved attributes (`locale`, `language`, `country`).
- Catalog-protocol distribution of code lists (`CatalogManifest`, `DependencyResolver`, `ImportCodeList`).

Follow-up issues raised during this PR:

- #391 — harden the URL fetch path (`Accept` header, max response size, content-type validation).
- #393 — variants UI: scale to many attributes by progressively adding one at a time instead of rendering every defined attribute as a row.
- #396 — cross-variant attribute validation (current per-variant validator is 1+K queries — fine for UI writes, deferred to land alongside bulk-write paths that carry code-list bindings).
- #397 — encrypt-at-rest for `code_lists.credential` and `catalogs.source_auth_credential` (both columns are plaintext today; addressed together since they share the same pattern).

## Test plan

- [x] `./gradlew :modules:epistola-core:test --tests "app.epistola.suite.attributes.codelists.*"` — 38 integration tests passing
- [x] `./gradlew :apps:epistola:uiTest --tests "app.epistola.suite.ui.CodeListUiTest"` — 8 Playwright tests passing
- [x] `./gradlew :modules:epistola-core:test --tests "app.epistola.suite.attributes.*"` — existing attribute tests still pass (no regression)
- [x] `./gradlew ktlintCheck` clean
- [x] `pnpm format:check` clean


apps/epistola/src/main/kotlin/app/epistola/suite/handlers/AttributeHandler.kt
```diff
@@ -93,20 +101,26 @@ class AttributeHandler {
...
+        val codeListSelection = request.params().getFirst("codeList")?.ifBlank { null }
+
+        val (allowedValues, codeListCatalogKey, codeListSlug) = parseConstraint(constraintKind, allowedValuesInput, codeListSelection)
 
         val result = form.executeOrFormError {
             CreateAttributeDefinition(
                 id = AttributeId(attributeKey, CatalogId(catalogKey, tenantId)),
                 displayName = displayName,
                 allowedValues = allowedValues,
+                codeListCatalogKey = codeListCatalogKey,
```
sdegroot: should this not be catalogId(key, slug) ?

apps/epistola/src/main/kotlin/app/epistola/suite/handlers/AttributeHandler.kt
```diff
@@ -173,14 +195,18 @@ class AttributeHandler {
...
-        val allowedValues = parseAllowedValues(allowedValuesInput)
+        val codeListSelection = request.params().getFirst("codeList")?.ifBlank { null }
+        val (allowedValues, codeListCatalogKey, codeListSlug) = parseConstraint(constraintKind, allowedValuesInput, codeListSelection)
 
         try {
             UpdateAttributeDefinition(
                 id = attributeId,
                 displayName = displayName,
                 allowedValues = allowedValues,
+                codeListCatalogKey = codeListCatalogKey,
```
sdegroot: also catalogId ?

apps/epistola/src/main/kotlin/app/epistola/suite/handlers/AttributeHandler.kt
sdegroot: Refactored in 322b5bcb — `CreateAttributeDefinition` (and `UpdateAttributeDefinition`) now takes a single `codeListId: CodeListId?` parameter instead of the two-key pair. The handler builds the `CodeListId` once in `parseConstraint`, and the storage shape stays the same in the DB (two columns) but the typed API surface is one composite. Added a same-tenant invariant check in the command init that wasn't possible to express with the old shape.

apps/epistola/src/main/kotlin/app/epistola/suite/handlers/AttributeHandler.kt
sdegroot: Same refactor in 322b5bcb covers the update path too — the line you flagged is gone, replaced by `codeListId = codeListId` constructed once from the form input.

modules/epistola-core/src/main/resources/db/migration/V29__code_lists.sql
```diff
@@ -0,0 +1,107 @@
...
+CREATE TABLE code_lists (
+    slug               CODE_LIST_KEY NOT NULL,
+    tenant_key         TENANT_KEY NOT NULL REFERENCES tenants(id) ON DELETE CASCADE,
+    catalog_key        CATALOG_KEY NOT NULL,
+    display_name       VARCHAR(100) NOT NULL,
+    description        TEXT,
+    source_type        CODE_LIST_SOURCE NOT NULL,
+    source_url         TEXT,
+    auth_type          CODE_LIST_AUTH_TYPE NOT NULL DEFAULT 'NONE',
+    credential         TEXT,
```
mvanleijenhorst, line 40: Is this field stored encrypted?

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/commands/variants/AttributeValidation.kt
```diff
@@ -21,10 +33,27 @@ fun validateAttributes(tenantId: TenantId, attributes: Map<String, String>) {
...
-            validate("attributes", value in definition.allowedValues) {
-                "Invalid value '$value' for attribute '$key'. Allowed values: ${definition.allowedValues.joinToString(", ")}"
-            }
+        validateValue(key, value, definition!!)
+    }
+}
+
+private fun validateValue(key: String, value: String, definition: VariantAttributeDefinition) {
+    val codeListId = definition.codeListId
+    if (codeListId != null) {
```
mvanleijenhorst, line 42: In AttributeValidation.kt:33 you have this:

  val definitions = ListAttributeDefinitions(tenantId).query()   // 1 query
  for ((key, value) in attributes) {
      ...
      val exists = CodeListEntryExists(...).query()              // N queries
  }

  The "1" is fine — all attribute definitions for the tenant in a single fetch. The "N" is the problem: for every
  code-list-bound attribute on the variant, a separate SELECT 1 round-trip hits Postgres. N = the number of 
  code-list-bound attributes on this single variant write.

  Important: N is not the number of code lists in the tenant, nor the number of entries in a list. It's strictly "how many
   fields on this variant point at a code list."

  When it's fine

  A typical editorial variant in your domain:

  attributes = {
    "language": "nl",
    "country":  "NL",
    "channel":  "print"
  }

  3 code-list fields → 1 + 3 = 4 queries for one variant. Unnoticeable; a local DB does that in under 5ms total. This is
  the happy path and will likely be ~95% of writes.

  When it goes wrong

  Scenario A — variants with lots of structured metadata. Suppose someone models a product variant with 12 standardised
  attributes (language, country, currency, unit_system, paper_size, locale, channel, customer_segment,
  distribution_region, tax_class, document_type, brand):

  1 + 12 = 13 round-trips per variant.write()

  Still fine on a local DB. On a prod DB with ~2ms round-trip latency: ~26ms of validation overhead per write. Noticeable
  on a single API call but livable.

  Scenario B — bulk import or seed. This is where it hurts. Imagine a template-publish or a CSV import of 5,000 variants,
  each with 12 code-list attributes:

  5,000 × 13 = 65,000 queries

  At 2ms round-trip that's 130 seconds of validation alone, where one well-shaped batch query would be under a second.
  This is the real risk: it scales with (variants) × (attributes per variant), so the moment anything bulk-shaped happens,
   it blows up.

  Scenario C — hot loops elsewhere. If validateAttributes ever ends up called per cell in a table render or per row in a
  preview (e.g. document-template variants validated per page), the multiplier compounds again.

modules/epistola-core/src/main/resources/db/migration/V29__code_lists.sql
sdegroot: its not, good point. ill make a new task for this. Its also with catalogs an issue

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/commands/variants/AttributeValidation.kt
sdegroot: Good catch — the math is right. I dug into where in the codebase the cliff actually surfaces and concluded the right move is to defer rather than batch within `validateAttributes`. Reasoning:

**The path that triggers the cliff doesn't ship in this PR.** The 5000-variant × 12-attr scenario is bulk writes through catalog import. Today catalogs can't carry code-list-bindings — the catalog-protocol extension for code lists is explicitly deferred in this PR's CHANGELOG ("Code lists — future work / Catalog-protocol distribution of tenant code lists"). Until that lands, the only path running `validateAttributes` against a code-list-bound attribute is UI writes, where K = 1-3 → 2-4 queries, rounding error.

**Per-variant batching is the wrong shape for the future bulk path.** `validateAttributes(tenantId, oneVariantAttributes)` operates on a single variant. Collapsing the K queries into 1 per variant helps single-variant wall-time at K ≥ 10 (which no real workflow today hits), but bulk writes would still pay `M × 2` queries — linear in M. The real fix for bulk is **cross-variant batching**: collect every `(codeListId, code)` across all M variants, dedupe (heavy overlap when many variants share `language=nl-NL` etc.), one query for the lot. That requires a different validator API — `validateAttributesForVariants` — called from the bulk-write loop, not from inside the per-variant command.

Filed #396 to track the cross-variant work, framed around the future call site ("land alongside catalog-protocol code-list distribution") rather than the current N+1 — so when that PR lands the validator can be batched in the same change with the right shape from day one.

modules/epistola-core/src/main/resources/db/migration/V29__code_lists.sql
sdegroot: Plaintext today — same shape as the existing `catalogs.source_auth_credential` column (introduced in V4), which we inherited the pattern from. Fixing this here for `code_lists.credential` only would leave that existing gap intact and create a divergence on what should be the same encrypt-at-rest pattern.

Filed #397 covering both columns together so the at-rest encryption layer ships once, consistently.

Issue #413, opened 2026-05-13
1.0: Normalise SQL migrations — collapse V1–V29 into V1__baseline.sql


sdegroot:
## Goal

Project is not yet in production (per CLAUDE.md). 1.0.0 is the moment the
migration history becomes append-only forever. Collapse before the freeze:

- Consolidate `modules/epistola-core/src/main/resources/db/migration/V1__app_metadata.sql`
  through `V29__code_lists.sql` into a single `V1__baseline.sql`.
- Decide a numbering convention for module-owned migrations (currently
  `loadtest` owns V11, `feedback` owns V17 in the same Flyway namespace).
  Options: numeric range allocation, or module-prefixed names like
  `V2__feedback_…`. Document in `docs/migrations.md`.
- Build the baseline from `pg_dump --schema-only` of an up-to-date DB;
  hand-crafting risks drift.

## Acceptance

- [ ] Fresh DB Flyway run reproduces a schema byte-identical to the current
      accumulated state (`pg_dump --schema-only` diff is empty).
- [ ] Integration tests pass against the baselined schema.
- [ ] `docs/migrations.md` published; CHANGELOG notes the one-time reset.

## Risk

This is breaking for existing dev databases — communicate as a one-time
reset.

Part of the 1.0.0 DoD #7.

PR #429, opened 2026-05-16
refactor!: standardize DB schema and wire audit columns

sdegroot:
Gated **pre-1.0, in-place** rewrite of the baseline migrations (analogous to #413) — every database is rebuilt from scratch; no production deployments, no migration paths. Makes the schema internally consistent, makes audit attribution a real enforced relationship, and closes the latent FK gaps that enforcement exposed.

## Commits

1. `refactor:` — pg_dump-equivalent normalization (TIMESTAMPTZ spelling, lowercase boolean literals, missing `COMMENT ON`). Build/tests unaffected.
2. `refactor!:` — the breaking schema standardization (renames, audit columns, types, domains).
3. `refactor:` — collapse per-purpose synthetic ids into one seeded `SystemUser`.
4. `fix:` — route test-user provisioning through the production `EnsureUser` command; drop raw test SQL; fixes the Playwright CI failure.
5. `fix:` — per-API-key service-account user (`AuthProvider.API_KEY`).
6. `refactor:` — `auth_providers` lookup table + DB-enforced `updated_at` triggers.

## What changed

**Naming (canonical):** `last_modified`→`updated_at`, `last_modified_by`→`updated_by`, `document_generation_requests.correlation_key`→`correlation_id` (matches `documents` + the REST contract). Threaded through entities, DAOs, MCP DTOs, handlers, Thymeleaf. External REST DTO field names kept (frozen contract — only mapper RHS changed).

**Audit columns wired end-to-end:** every mutable table carries `created_by`+`updated_by`, populated from the acting principal and mapped back onto the entity. All audit FKs to `users(id)` are `ON DELETE SET NULL` (deleting a user nulls attribution, doesn't cascade the row). Deliberate exception: `feedback.created_by` stays mandatory `NOT NULL`. `updated_at` is now **DB-enforced** by a shared `set_updated_at()` `BEFORE UPDATE` trigger on every table that has it, so the timestamp can't rot if a DAO forgets it.

**Types / consistency:** `themes.spacing_unit` `REAL`→`NUMERIC(6,2)` (Kotlin stays `Float?`; JDBI converts). All timestamps `TIMESTAMPTZ`; lowercase boolean literals; `TENANT_KEY` domain replaces raw `VARCHAR`/`TEXT` tenant columns in `event_log`/`generation_results`/consumer/feedback tables. Missing table/column comments added. Spring Session tables untouched.

**Every writer is a real `users` row** (the point of enforcing the FK):
- One canonical **`SystemUser`** (all-zeros UUID) **seeded by the `core_users` baseline migration** — a DB invariant like `installation` — used by the job poller, demo bootstrap, demo login resolver, and system-catalog install/upgrade. Replaces the previous mix of per-purpose synthetic ids and the runtime "ensure system user" hack; background-generated documents now attribute to `SystemUser` instead of `NULL`.
- **Latent production gaps fixed** (FK enforcement exposed them): `LocalUserDetailsService` (config users) and `ApiKeyAuthenticationFilter` now provision a `users` row via the idempotent `EnsureUser` command on auth — so local-auth and `X-API-Key` REST writes attribute precisely and don't FK-violate. Each API key is its own service identity (new `AuthProvider.API_KEY`).
- `EnsureUser` is the single idempotent provisioning primitive (atomic, target-less `ON CONFLICT DO NOTHING`) shared by production auth **and** the test harness — no test-only SQL, no fixed seed list.
- `users.provider` is a FK to a new **`auth_providers`** lookup table (extend by inserting a row, no enum/schema change), created before `users` so the FK is inline (no `ALTER`).

**Test infrastructure:** `TestPrincipalUsers` seam materialises a `users` row for whatever principal the harness binds (via `EnsureUser`); bespoke raw-SQL `@BeforeAll` seeds removed from `ApiKeyHandlerTest`/`FeedbackIntegrationTest` (review comment); `DemoLoader`'s redundant user/membership loop removed; `AuditColumnsIT` + an `ApiKeyAuthenticationFilter` regression test added.

## Verification

- `./gradlew unitTest integrationTest` ✅ · full `./gradlew test` incl. Playwright ✅ (the previously-failing UI suite run repeatedly clean)
- `./gradlew ktlintCheck` ✅ · `pnpm format:check` ✅
- `docs/migrations.md` + `CHANGELOG.md` updated. No demo-catalog resources changed → no version bump.

## Out of scope (tracked)

Deferred pre-1.0 hardening → #430 (DEFAULT partitions, `content_store` GC, secret encryption, `event_log` retention, FK `ON DELETE` audit, drift guardrail, removing the last cross-file deferred `ALTER`s). Separating DB migration from app startup → #431.

## Reviewer notes

One-time effect: existing local dev DBs auto-clean + rebuild on next start; demo data reloads. The audit-wiring across command handlers is mechanical and consistent (`auditUser` local, `:createdBy`/`:updatedBy` binds).


apps/epistola/src/test/kotlin/app/epistola/suite/feedback/FeedbackIntegrationTest.kt
```diff
@@ -53,7 +53,7 @@ class FeedbackIntegrationTest : BaseIntegrationTest() {
                 """
                 INSERT INTO users (id, external_id, email, display_name, provider, enabled, created_at)
```
sdegroot: why do we have sql statements in the test like this?

PR #435, opened 2026-05-17
feat(templates): backend JSONata syntax validation for stencil parameter bindings (#385)

### LINKED ISSUE #385: Stencil parameters: backend JSONata syntax validation on save
## Context

`NodeParameterBindingValidator` checks that bindings reference declared parameter names and that required parameters are bound, but it does **not** validate that the bound expression is syntactically valid JSONata. A typo like `recipient.firstName & '` (unbalanced quote) saves cleanly and only blows up at render time.

## Scope

- Plug the JSONata library (`com.dashjoin:jsonata`, already on the classpath in `:modules:generation`) into the validator path. Pre-compiling each expression at validation time would catch parse errors.
- Surface a new error code, e.g. `NODE_PARAMETER_BINDING_SYNTAX_INVALID`, with the JSONata error message attached.
- Beware: `:modules:epistola-core` doesn't currently depend on the JSONata library (it lives in `:modules:generation`). The validator may need to move or a small wrapper added to keep the dependency direction clean.

## References

- Tracked in `docs/stencil-parameters.md` §9
- Decision context: ADR 0002 D1 (lazy evaluation)
- Feature toggle: `stencil-parameters` (default off)

whit3st:
## Description

Backend JSONata syntax validation for stencil parameter bindings — **plus** a first-class validation error-code model that this work grew into (folded in deliberately during review).

### 1. JSONata syntax validation (original scope, #385)

`NodeParameterBindingValidator` now pre-compiles every binding expression with `com.dashjoin:jsonata` at save time (parse-only, no evaluation), catching syntax errors before they reach the renderer. The template draft-save route returns these as a structured `400` instead of a generic `500`; the editor validates inline and disables Save on known-invalid JSONata. Explicit `com.dashjoin:jsonata` dependency added to `:modules:epistola-core` (it was only a transitive `implementation` of `:modules:generation`).

### 2. First-class validation error codes (scope expansion)

The ~21 domain codes (`NODE_PARAMETER_BINDING_*`, `PARAMETER_*`, `PLACEHOLDER_*`, `STENCIL_RECURSION`, `PAGEHEADER_*`) used to be `SCREAMING_CODE:` prefixes inside `ValidationException.message`, forcing the editor to recover them by regex. They are now a first-class `code: ValidationCode` field:

- `ValidationException` gains `code: ValidationCode = GENERIC` (wire `VALIDATION_ERROR`) — the ~80 non-coded sites are unchanged via the default; only the 21 coded throw sites were migrated, messages are now human-only.
- A single shared mapper `ValidationException.toValidationErrorResponse()` is used by **both** the REST `ApiExceptionHandler` and the UI draft-save route, so the specific code is reported consistently (REST previously flattened everything to a generic `VALIDATION_ERROR`).
- The editor consumes the structured `code`/`field` (new `binding-errors.ts`); the `parseBindingSaveError` regex is gone. The editor↔inspector channel now shares a typed `STENCIL_BINDING_ERRORS_KEY` and a pure, unit-tested `bindingErrorsForSaveState` helper.
- This mapper is the single documented seam for the planned **RFC 7807 `application/problem+json`** migration (not implemented here; wire shape and `epistola-contract` OpenAPI unchanged).

### 3. Review findings addressed

- Blank/empty required binding now reports the precise `NODE_PARAMETER_BINDING_MISSING_REQUIRED` instead of a misleading syntax error.
- `UiExceptionFilter` error body is now Jackson-encoded (pre-existing bug: hand-built `{"error":"$message"}` broke on quotes/backslashes/newlines).
- Deferred `validate()` in `binding-row.ts` is guarded against running after dialog teardown.
- Dual-parser divergence (Java `com.dashjoin:jsonata` vs npm `jsonata`) documented as an accepted trade-off.

## Related Issue(s)

Closes #385

Follow-up (review finding 6, scope gap): a separate issue tracks centralizing `ValidationException → 400` for **all** UI handlers (only the draft-save route is fixed here, on purpose; the `UiExceptionFilter` JSON fix in this PR makes that future change safe).

## Type of Change

- [x] Documentation update
- [x] Refactoring (no functional changes)
- [x] Chore (dependencies, CI, tooling)

## Component(s) Affected

- [x] Documentation
- [ ] CI/CD

## Checklist

- [x] New and existing tests pass locally

## Testing

- `./gradlew :modules:epistola-core:unitTest :modules:rest-api:test` — green
- `./gradlew :modules:epistola-core:integrationTest --tests '*StencilPlaceholderIntegrationTest'` — green
- `./gradlew :apps:epistola:integrationTest --tests '*DocumentTemplateRoutesTest'` — green (incl. `DraftSaveEndpointTest`)
- `./gradlew :apps:epistola:test --tests 'app.epistola.suite.config.UiExceptionFilterTest'` — green
- `pnpm --filter @epistola/editor test` — 1712 pass; `tsc -b` clean
- `./gradlew ktlintCheck` + `pnpm format:check` — clean

## Additional Notes

- Validation is syntax-only; it does not evaluate expressions or check data-model field existence / declared types.
- Pre-production: no backward-compat shims (per project policy) — the ~21 coded throw sites were migrated directly.
- Out of scope (separate epic): folding the 30+ hardcoded `ApiExceptionHandler` codes into `ValidationCode`; the actual RFC 7807 body/media-type switch.


modules/epistola-core/build.gradle.kts
```diff
@@ -42,6 +42,10 @@ dependencies {
     // Validation
     implementation(libs.json.schema.validator)
 
+    // JSONata expression language (already in :modules:generation; declared here
+    // for compile-time visibility in epistola-core validators)
+    implementation("com.dashjoin:jsonata:0.9.10")
+
```
whit3st, line 48: I should've created a libs alias here

sdegroot:
Sounds like a reasonable thing, though I wonder: do we also check for this in the frontend? And what hapepns if this error occures?

whit3st:
> Sounds like a reasonable thing, though I wonder: do we also check for this in the frontend? And what hapepns if this error occures?

It was purely a backend validation, now the errors are shown in the UI, you were right to question it. I will do a one final check tomorrow. WIP now

whit3st:
@sdegroot I think this is ready. Per issue description, this PR adds JSONata syntax checks, nothing more and I wonder if we can/should do:

1. Field existence validation. e.g. `customer.name` is valid but I dont know how to proceed if variable(s) is/are used in JSONata. There are thousand of ways to use a variable?
2. Result type validation. parameter declared as `boolean` but expression evaluates to `'true'` which can be done easily?
3. Probably example-data evaluation as well. Evaluate bindings against available data examples and report runtime failures early maybe? Everything except syntax error can be a warning, not a blocker. Backend is still source of truth. 

Let me know any of these are worth pursuing or maybe later.

PR #448, opened 2026-05-19
feat(catalog): versioning Phase A/2 — upgrade preview, ZIP-exchange semantics, drift hint

sdegroot:
## Summary

Rounds out catalog import/export into a coherent, well-tested feature. Builds on the Phase 1 release/fingerprint model already on `main`.

### Subscriber upgrades (Phase 2)
- **`PreviewCatalogUpgrade`** (`TENANT_SETTINGS`): source-vs-source `UpgradeDiff` (added/removed/changed/unchanged + blocking conflicts) using a stored `catalogs.installed_resource_fingerprints` baseline, so CHANGED means *the publisher* changed it — never install round-trip noise. Conflict/stale logic extracted into a shared `CatalogUpgradeAnalyzer` used by both preview and apply.
- **UI**: explicit per-row **refresh icon** (never auto-polls); result rendered **in the Version column** (green `vX` up-to-date / greyed `→ vY` update / `vX · ZIP` mirror / `vX · check failed`). Review dialog (CSP-safe, `hx-get` only) with opt-in "also install new resources" (`UpgradeCatalog.includeNewSlugs`) and server-rendered disabled Apply on conflicts.
- **Read parity**: `GET /api/tenants/{tenantId}/catalogs/{catalogId}/upgrade-preview` + MCP `preview_catalog_upgrade`. Upgrade *action* stays UI-only (mirrors the release-action decision).

### Upgrade correctness (Phase A)
- `UpgradeCatalog` **aborts before any destructive change** on a failed resource install (no silent half-upgrade; `aborted` flag; reconciler maps to a loud, self-retrying failure). `EnsureSubscribedCatalog` no longer re-upgrades every boot on a null-fingerprint manifest.

### ZIP exchange semantics
- ZIP import of a **SUBSCRIBED** catalog is now a full `UpgradeCatalog` (conflict-check, prune, abort-on-fail).
- ZIP import that **creates** an AUTHORED catalog adopts the manifest release as its initial release; importing into an **existing** AUTHORED catalog offers **Merge vs Replace** (`authoredMode`), Replace conflict-checked before mutating. Release state never fabricated.
- `CatalogZipVersioningTest` pins the semantics.

### UI hardening
- Resilient browse (ZIP-managed / unreachable SUBSCRIBED degrades to installed resources instead of 500), stable parser-safe `#catalog-list` swap region, radio/checkbox sizing, code-list/font shown as installed + type badges.

### AUTHORED working-copy drift hint + B2 refactor
- The list shows `vX · pending changes` when an AUTHORED released catalog's working copy drifted (cheap timestamp baseline `GREATEST(released_at, imported_at)`, new `catalogs.imported_at` so a no-op re-import isn't drift).
- Final shape (`02cc326d`): a dedicated `ListCatalogsForManagement` query → `CatalogListRow` (`@Nested Catalog` + list-only `pendingChanges`), drift in one `LEFT JOIN`. Shared `Catalog` read model untouched; no parallel id-set / SpEL join.
- **Fix surfaced by the above**: `ReleaseCatalogVersion` now stamps `released_at` from the DB clock (`SELECT NOW()`), not JVM time — eliminates a JVM↔DB clock-skew false "pending changes".

## Testing
All catalog suites green via Testcontainers: `ListCatalogsForManagementTest`, `CatalogListHandlerTest`, `CatalogReleaseHandlerTest`, `CatalogUpgradeHandlerTest`, `PreviewCatalogUpgradeTest`, `CatalogZipVersioningTest`, `ExportUsesReleasedVersionTest`, `CatalogExportImportTest`, `UpgradeCatalogTest`. ktlint + `pnpm format:check` clean. CHANGELOG + `docs/catalog-versioning.md` updated.

## ✅ Contract gate (resolved)
The REST read-parity commit (`d9725774`) consumes **`epistola-contract` 0.5.2** (`previewCatalogUpgrade` + `CatalogUpgradeDiff`), released from epistola-contract PR #26. `epistola-contract` 0.5.2 is now **published**; `gradle/libs.versions.toml` pins it and `origin/main` is merged in cleanly, so CI now builds the full branch against 0.5.2.


modules/rest-api/src/main/kotlin/app/epistola/suite/api/v1/EpistolaCatalogApi.kt
```diff
@@ -40,6 +44,33 @@ class EpistolaCatalogApi : CatalogsApi {
...
     }
 
+    override fun previewCatalogUpgrade(
+        tenantId: String,
+        catalogId: String,
+    ): ResponseEntity<CatalogUpgradeDiff> {
+        val diff = PreviewCatalogUpgrade(
+            tenantKey = TenantKey.of(tenantId),
+            catalogKey = CatalogKey.of(catalogId),
+        ).query()
```
mvanleijenhorst, line 55: **1. REST returns 500 on a non-subscribed catalog (and the MCP tool surfaces a generic error).** `PreviewCatalogUpgrade.handle` throws `IllegalStateException("…only subscribed catalogs can be upgraded")` for AUTHORED/no-source-URL catalogs, and `IllegalArgumentException("Catalog not found: …")` for unknown slugs — both bubble as 500 here. The UI handler (`CatalogHandler.upgradeDialog`) catches and re-renders the dialog with an error alert; this surface doesn't.

Wrap the `.query()` call in a try/catch and map to 404 (not found) / 409 (wrong catalog type) / 422 as appropriate. `CatalogMcpTools.previewCatalogUpgrade` has the same issue and should mirror the mapping.

apps/epistola/src/main/kotlin/app/epistola/suite/handlers/CatalogHandler.kt
```diff
@@ -242,6 +240,104 @@ class CatalogHandler {
         }
     }
 
+    /**
+     * Lazy per-row "upgrade available?" indicator (SUBSCRIBED only). Cheap —
+     * manifest fetch only. Renders nothing when up to date or unreachable, so a
+     * failed remote never breaks the list.
+     */
```
mvanleijenhorst: **4. Orphaned KDoc — describes the prior auto-polling behaviour, not the current handler.** Lines 243–247 say "Lazy per-row indicator … Renders nothing when up to date or unreachable" — but the explicit Check-for-updates change in `4fe23a15` made this handler render the `version-status` fragment in *every* state (UP_TO_DATE / UPDATE_AVAILABLE / ZIP_MANAGED / CHECK_FAILED), and `CatalogUpgradeHandlerTest` asserts the list never auto-polls (`doesNotContain("hx-trigger=\"load\"")`).

Kotlin attaches only the **last** preceding KDoc to a declaration, so this block is dead text — but it actively misleads readers ("which of these two docs is right?"). The correct KDoc immediately follows on lines 248–255.

```suggestion
```

mvanleijenhorst (commented):
Four polish points before merge — architecture / tests / docs all look good. Inline comments below for items 1 and 4 (in-diff). Items 2 and 3 follow as a conversation comment because `importCatalog` itself isn't touched by this PR.

mvanleijenhorst:
Two more items on `modules/rest-api/.../EpistolaCatalogApi.kt` — both on `importCatalog`, which this PR doesn't touch directly but which is now out of sync with the new SUBSCRIBED-ZIP behaviour the PR introduces. Posting as a conversation comment since the relevant lines aren't in the diff.

---

**2. Cross-surface drift — `authoredMode` is UI-only.** `ImportCatalogZip` accepts `authoredMode: AuthoredImportMode` and the Import-ZIP dialog exposes Merge/Replace radios, but REST silently defaults to MERGE: `EpistolaCatalogApi.importCatalog(tenantId, file, catalogType)` never reads the parameter, so the new Replace behaviour is unreachable from the REST surface.

Per `CLAUDE.md` item 11 ("Consider all surfaces"), either:

- expose `authoredMode` here (add an optional form/query parameter in `epistola-contract` 0.5.3, default `MERGE`), or
- land an explicit decision note (CHANGELOG + KDoc on `ImportCatalogZip.authoredMode`) that Merge/Replace is a UI-only convenience and REST is always MERGE.

---

**3. Response drops the `aborted` signal — API clients can't distinguish abort-and-retry from partial success.** `ImportCatalogZipResult.aborted = true` is a distinct outcome of a SUBSCRIBED ZIP upgrade introduced by this PR: one or more installs failed, so stale resources were **not** pruned and `installed_*` was **not** advanced — the catalog is still on its previous release and a re-import is meaningful. A non-aborted result with `failed > 0` means the failures are permanent for that import and the catalog moved forward.

Today a client gets `{ installed: 0, updated: 3, failed: 1 }` for both cases and has no way to tell them apart. A nightly sync script can't tell "the publisher shipped a broken ZIP, retry tomorrow" from "the publisher made breaking changes, escalate."

Add `aborted: Boolean` to `ImportCatalogResponse` in `epistola-contract` (next minor, default `false`) and surface `result.aborted` in `EpistolaCatalogApi.importCatalog`. The MCP `import_catalog` tool (if/when it lands) should carry the same field.

sdegroot:
Thanks @mvanleijenhorst — all four addressed.

**1. REST 500 / MCP generic on non-subscribed/unknown catalog** — fixed in `4c6e4462`. `PreviewCatalogUpgrade` now throws typed `CatalogNotFoundException` and `CatalogNotUpgradeableException`; `ApiExceptionHandler` maps them to **404** / **409** (was a misleading 400 / 500). MCP surfaces the clear typed message instead of a generic error (no HTTP status there). Two negative tests added — `PreviewCatalogUpgradeTest` 5 → 7.

**4. Orphaned KDoc (`CatalogHandler.kt:243–247`)** — fixed in `4c6e4462`. The stale auto-poll KDoc block is deleted; the correct KDoc on `upgradeCheck` remains.

**2. `authoredMode` UI-only on REST import** & **3. `aborted` dropped from `ImportCatalogResponse`** — chose **full parity** (not the doc-only option). Contract change: **epistola-contract PR https://github.com/epistola-app/epistola-contract/pull/27** (`0.5.2 → 0.5.3`) adds the optional `authoredMode` form field (`MERGE`|`REPLACE`, default `MERGE`) and required `ImportCatalogResponse.aborted`. Suite side wired in `2a91dba1`: `EpistolaCatalogApi.importCatalog` reads `authoredMode` → `ImportCatalogZip`, and maps `result.aborted` into the response. There is **no MCP `import_catalog` tool yet**, so import is not an MCP surface (noted in CHANGELOG/docs); when it lands it'll carry the same field.

CHANGELOG + `docs/catalog-versioning.md` updated for all of the above. Note: as with `0.5.2 → previewCatalogUpgrade`, this branch's CI is red until **epistola-contract 0.5.3 is published** (contract-first flow); it goes green on publish. #1/#4 are independently green (`PreviewCatalogUpgradeTest` 7, handlers 6/3).

PR #451, opened 2026-05-19
fix(generation): preserve empty paragraphs as blank lines in PDF

### LINKED ISSUE #400: [Bug]: Enter / line breaks in text component appear to be dropped
## Summary

When pressing Enter inside a `text` component, the resulting line break / paragraph break does not appear to make it through to the rendered output. The exact scenarios where this fails still need to be reproduced and documented — this issue tracks the investigation as much as the fix.

## Steps to reproduce (to be verified)

1. Open the template editor.
2. Add a `text` component.
3. Type some text, press Enter, type more text.
4. Generate / preview the document (HTML preview and PDF).
5. Observe whether the line break / new paragraph is preserved.

## Expected behaviour

Pressing Enter inside the text component produces a visible break in the rendered output — either a new paragraph or a hard line break, consistent with what the editor shows.

## Actual behaviour

Breaks appear to be lost somewhere between the editor and the rendered output. Needs confirmation on exactly where (editor state, persisted model, HTML render, PDF render) the break disappears.

## Investigation checklist

- [ ] Reproduce on a clean template.
- [ ] Confirm whether the break is preserved in the persisted ProseMirror/TipTap document (editor model).
- [ ] Confirm whether the break is preserved in the HTML preview.
- [ ] Confirm whether the break is preserved in the generated PDF.
- [ ] Test both \"soft break\" (Shift+Enter / `<br>`) and \"hard break\" (Enter / new paragraph).
- [ ] Test in different text-bearing components if more than one exists (e.g. paragraph, heading, list item).
- [ ] Test with inline formatting present around the break (bold, italic, links, expressions).
- [ ] Test with empty paragraphs (Enter on an empty line).
- [ ] Test multiple consecutive breaks.
- [ ] Document the exact scenarios that fail vs. work.

## Notes

- May be related to the architectural concerns raised in #312 (multi-paragraph text blocks with their own rendering path via `TipTapConverter`).
- Once scenarios are confirmed, this issue may need to be split into separate bugs per failure mode.

mvanleijenhorst:
## Summary

- Pressing Enter twice in the template `text` editor inserts an empty `paragraph` node — the editor shows a blank line, but the rendered PDF dropped it because `iText.Paragraph` with no inline content collapses to zero height.
- `ProseMirrorConverter.convertParagraph` / `convertHeading` now insert a non-breaking space (`U+00A0`) into any empty segment, preserving one line of vertical space.
- Covers both empty paragraph nodes and blank segments produced by a `hard_break` at the start, middle, or end of a paragraph.

## Scope of #400

The issue is an investigation umbrella. This PR addresses the **empty-paragraph case** that the reporter confirmed in conversation ("twee keer Enter → tekst, lege regel, tekst — wel in editor, niet in PDF"). Other items on the issue checklist (soft vs. hard break in all components, marks around breaks, multiple consecutive breaks elsewhere) are covered by the new tests for `text` components specifically, or remain as separate follow-ups if they surface.

## Test plan

- [x] New `ProseMirrorConverterTest`: `empty paragraph between filled paragraphs reserves a blank line` (asserts placeholder content on the empty segment).
- [x] New `ProseMirrorConverterTest`: `paragraph with only a hard break produces two blank-line segments`.
- [x] New `TextNodeMarginRegressionTest`: `empty paragraph between two filled paragraphs renders as a visible blank line` (real-PDF baseline-gap measurement; verifies the gap grows by ≥ 8pt when the empty paragraph is added).
- [x] `./gradlew :modules:generation:test` — all 40+ tests green; no PDF baseline regressions.
- [x] `./gradlew :modules:generation:ktlintCheck` — clean.
- [x] `pnpm format` — no diff outside of these files.

Closes #400


sdegroot:
@mvanleijenhorst the build seems to be failing, this is why i have not started reviewing yet

mvanleijenhorst:
> @mvanleijenhorst the build seems to be failing, this is why i have not started reviewing yet

Build if fixed

PR #459, opened 2026-05-20
feat!: themes are optional; ship system catalog default theme

sdegroot:
## Summary

- **Themes are now optional.** `CreateTenant` no longer auto-creates a "Tenant Default" theme or populates `tenants.default_theme_key`. Templates without a theme render with the engine defaults already in place (A4 portrait, 20mm margins, Helvetica fallback) — no user-visible degradation.
- **Bundled `system` catalog ships a `default` theme** (Inter, 11pt, A4) that any tenant can opt into as its tenant default. Existing tenants pick it up on next boot via `SystemCatalogBootstrap`'s fingerprint-driven upgrade.
- **`SetTenantDefaultTheme` accepts a nullable `themeId`** to clear the tenant default and now correctly writes `default_theme_catalog_key` alongside `default_theme_key`. UI gains a "Clear default theme" action on the active default row.
- **`LastThemeException` is removed.** Zero themes is a valid state. `DeleteTheme` still refuses to delete the current tenant default — clear it first. The REST `LAST_THEME` code and the UI `#lists.size(themes) > 1` guard are dropped.

## Why

The auto-created "Tenant Default" theme existed for UX consistency ("here's a theme row in the list you can edit"), not technical necessity. Every fallback layer already handles a null theme: `ThemeStyleResolver` → `RenderingDefaults.V1` → `StyleApplicator` font fallback → editor `DEFAULT_PAGE_SETTINGS`. Forcing every tenant to own a magical theme they didn't ask for is friction without benefit.

## Future work

This PR keeps things minimal. Two natural follow-ups, deliberately out of scope:

- **Tenant setting to pick a default theme from any installed catalog.** Today the only way to set the tenant default is via the themes-list star icon (per-theme). A dedicated "Default theme" picker in tenant settings — listing every theme across all installed catalogs (system, demo, authored, subscribed) with "None" as a first-class option — would be the natural home for this and would integrate cleanly with the now-nullable `SetTenantDefaultTheme`.
- **A migration to clean up legacy auto-created themes** in long-lived deployments (deliberately skipped here since we're pre-production).

## Breaking change

New tenants no longer have a default theme until one is selected. Existing tenants are unaffected — their previously auto-created "Tenant Default" themes remain as regular user-owned themes that can be deleted or kept.

## Test plan

- [x] `./gradlew unitTest` (incl. `BundledCatalogFingerprintTest` — system catalog manifest fingerprint regenerated)
- [x] `./gradlew integrationTest` (incl. updated `TenantCommandsTest` asserting `defaultThemeKey == null` post-create; `SystemFontsIntegrationTest` rerouted to explicitly use `system/default` so live-render font-embedding still has Inter to find)
- [x] `./gradlew ktlintCheck` + `pnpm format:check`
- [ ] Manual smoke: create new tenant → themes list shows only `system/default` (read-only, from system catalog); template without theme renders A4/20mm/Helvetica; star `system/default` → it becomes tenant default; click "Clear default" → tenant has no default again


apps/epistola/src/main/resources/templates/themes/list.html
```diff
@@ -75,24 +76,34 @@ <h1>Themes</h1>
                         </td>
                         <td th:text="${#temporals.format(theme.updatedAt, 'yyyy-MM-dd HH:mm')}"></td>
                         <td class="actions">
-                            <span th:if="${theme.catalogType.name() == 'SUBSCRIBED'}" class="badge badge-muted" title="Subscribed catalogs are read-only">Read-only</span>
+                            <span th:if="${theme.catalogType.name() == 'SUBSCRIBED' and !isDefault}" class="badge badge-muted" title="Subscribed catalogs are read-only">Read-only</span>
```
mvanleijenhorst: The `and !isDefault` clause hides the "Read-only" badge in a state that this PR makes reachable for the first time: a subscribed theme (e.g. `system/default`) selected as the tenant default. State matrix:

| catalogType | isDefault | Badges shown next to name |
|-------------|-----------|---------------------------|
| AUTHORED    | false     | —                         |
| AUTHORED    | true      | **Default**               |
| SUBSCRIBED  | false     | **Read-only**             |
| SUBSCRIBED  | true      | **Default** (read-only signal lost) |

In that last row, the only remaining cue that the theme isn't editable is the *absence* of an edit button — an indirect signal at best.

Simplest fix is to drop the `!isDefault` guard:

```html
<span th:if="${theme.catalogType.name() == 'SUBSCRIBED'}"
      class="badge badge-muted" title="Subscribed catalogs are read-only">Read-only</span>
```

Two badges side by side ("Default Read-only") read fine and stay informationally correct.

mvanleijenhorst (commented):
Three follow-ups from the deeper review pass. Treating these as advisory rather than blocking — author's call on which to address before merge.

---

### 1. Missing tests for the clear-default path

**File:** `modules/epistola-core/src/test/kotlin/app/epistola/suite/tenants/TenantCommandsTest.kt`

The new "themes are optional" semantics rest on three branches of `SetTenantDefaultTheme` that are not yet covered:

1. `themeId = null` clears both `default_theme_key` and `default_theme_catalog_key`.
2. `themeId != null` with a `catalogKey` that doesn't contain that theme throws `ThemeNotFoundException`.
3. `DeleteTheme` succeeds on a theme after the tenant default has been cleared (the old `LastThemeException` path is gone — worth pinning that down explicitly so it doesn't quietly come back via a `COUNT(*) <= 1` guard somewhere).

Suggested additions (matching the existing `given/whenever/then` style):

```kotlin
@Test
fun `SetTenantDefaultTheme with null themeId clears the default`() = fixture { ... }

@Test
fun `SetTenantDefaultTheme rejects a themeId that does not exist in the given catalog`() = fixture { ... }

@Test
fun `DeleteTheme succeeds after the tenant default has been cleared`() = fixture { ... }
```

The `CreateTenant does not assign a default theme` test you added is great for the create-side invariant; these would round out the set-side and delete-side.

---

### 2. Dead code left behind by `LastThemeException` removal

**File:** `modules/epistola-core/src/main/kotlin/app/epistola/suite/themes/queries/CountThemesForTenant.kt`

`git grep` for callers on this branch turns up nothing — `DeleteTheme` was the only consumer, and the count check is gone now. The KDoc still says:

```kotlin
/**
 * Counts the number of themes for a tenant.
 * Used to prevent deleting the last theme.
 */
```

Two options:

- **Delete it.** Cleanest — the query, its handler, and its `THEME_VIEW` permission binding all go away.
- **Keep it, but retarget the docstring.** Only if the upcoming tenant-settings "default theme picker" mentioned in the PR description is expected to need a theme count. In that case, drop the "prevent deleting" line so the comment doesn't lie about its purpose.

Either is fine; leaving it as-is is the only option I'd push back on.

---

### 3. Manual smoke checkbox still unchecked

In the test plan:

```
- [ ] Manual smoke: create new tenant → themes list shows only `system/default` …
```

For a `feat!` PR that flips a long-standing invariant and changes which catalog types can become the tenant default, this is worth running before merge. The UI path exercises three things no automated test currently covers end-to-end:

- The "Clear default theme" HTMX form (new template fragment, new handler branch).
- Setting a **subscribed** theme (`system/default`) as default — this combination wasn't reachable before this PR.
- Re-render of the row after the HTMX swap: the badge/button state for the four `catalogType × isDefault` combinations (see the inline comment on `list.html`).

Happy to do the smoke run if it's blocking — let me know.

sdegroot:
Thanks for the review — addressed three of four in 4785d960:

1. **Tests for clear/reject/delete paths** — added the three suggested tests to `TenantCommandsTest.kt`. All three branches now pinned: `null themeId` clears both key + catalog key; non-existent themeId throws `ThemeNotFoundException`; `DeleteTheme` succeeds on a theme once the tenant default has been cleared (the dead `COUNT(*) <= 1` guard can't quietly return now).
2. **Dead `CountThemesForTenant` query** — deleted. It was the only consumer of the removed last-theme guard; leaving a public query bound to `THEME_VIEW` would have invited drift.
3. **Inline `!isDefault` guard on the Read-only badge** — dropped. Both "Default" and "Read-only" now render side-by-side for a subscribed theme that is the tenant default — the previously indirect cue (absence of an edit button) is now explicit.
4. **Manual smoke** — happy for you to do it if you've got the test env warm; otherwise I'll run through it before merge.

mvanleijenhorst (approved):

PR #467, opened 2026-05-21
fix(api): scope catalog-scoped list endpoints to the requested catalog (#466)

### LINKED ISSUE #466: listTemplates ignores catalogId path parameter and returns all tenant templates
## Problem

`GET /api/tenants/{tenantId}/catalogs/{catalogId}/templates` ignores the `catalogId` path segment and returns every template for the tenant. The same response body is returned for every catalog — including catalogs that do not own the listed templates.

Downstream, this causes 404 cascades in the Valtimo plugin's generate-document configurator: the user picks a catalog, the dropdown is populated from this endpoint with templates that don't belong to that catalog, and the next call (`getTemplate(tenantId, catalogId, templateId)`) returns 404 because the template really isn't in that catalog.

### Concrete example

Demo tenant has four catalogs (`default`, `epistola-demo`, `municipality-demo`, `system`). Probing each of the two non-system, non-default catalogs returns identical responses:

```bash
$ curl -s -H 'X-API-Key: epk_demo_…' -H 'Accept: application/vnd.epistola.v1+json' \
    http://localhost:4000/api/tenants/demo/catalogs/epistola-demo/templates \
  | jq -r '.items[].id' | sort
belastingbrief
besluit-bezwaar
bevestigingsbrief-vergunning
demo-invoice
example-template
financieel-overzicht
hello-world
ontvangstbevestiging-bezwaar
simple-letter
subsidiebesluit
test
voorwaarden-bijlage
xxx

$ curl -s -H 'X-API-Key: epk_demo_…' -H 'Accept: application/vnd.epistola.v1+json' \
    http://localhost:4000/api/tenants/demo/catalogs/municipality-demo/templates \
  | jq -r '.items[].id' | sort
# identical 13 items, including ones authored under `default` (e.g. xxx, test,
# simple-letter, hello-world, demo-invoice) that are not in either classpath catalog
```

Each item's `lastModified` timestamp also matches across catalogs, so this is the same data, not coincidentally identical results.

The `TemplateDto` returned by the endpoint doesn't include a `catalogId` field either, so clients can't filter the response themselves.

## Expected behaviour

- `listTemplates(tenantId, catalogId)` returns only templates that belong to the named catalog (a tenant catalog must be a partition of its templates, otherwise `catalogId` in the path is meaningless).
- The response items either include `catalogId` explicitly or `catalogId` remains purely a server-side filter — either way, the user/client should never see templates from other catalogs in a per-catalog listing.
- A request for templates in a non-existent catalog returns `404 Not Found` (already covered by the standard catalog-existence check, but worth confirming on this code path).

## Scope

- [ ] Audit the `listTemplates` query path in the templates module — confirm whether the SQL/JPA filter on `catalogId` is missing, joining the wrong column, or being short-circuited.
- [ ] Decide whether `TemplateDto` should grow a `catalogId` field for round-trip clarity (every other list-style DTO in the API does this).
- [ ] Add an integration test: seed templates in two catalogs and assert each catalog's listing only contains its own templates.
- [ ] Apply the same audit to any sibling endpoints that take `catalogId` in the path but might similarly ignore it (variants, attributes, code lists, fonts, stencils).
- [ ] If `TemplateDto` does NOT carry `catalogId`, also confirm that `getTemplate(tenantId, catalogId, templateId)` enforces the catalog binding — i.e. asking for `catalogs/epistola-demo/templates/subsidiebesluit` when subsidiebesluit lives in `municipality-demo` should return 404, not the wrong template.

## Notes

Discovered while debugging a 404 cascade in `valtimo-epistola-plugin` (`generate-document` process-link configurator) — when a user switched catalogs in the dropdown, the new templates list was populated with the wrong templates and the next API call 404'd. The plugin side has a separate UX fix landing for the stale-`v-select`-state half of the bug (`valtimo-epistola-plugin@caa038e`), but the underlying cross-catalog leak is here.

Tested against Epistola Suite running locally on `localhost:4000` from the latest `main` (`epistola-suite` ≥ 0.5.x, contract `client-spring3-restclient` 0.6.0) on 2026-05-21.

sdegroot:
## Summary

`GET /api/tenants/{tenantId}/catalogs/{catalogId}/templates` and `GET /api/tenants/{tenantId}/catalogs/{catalogId}/stencils` were silently dropping the `catalogId` path parameter and returning every template/stencil owned by the tenant across **all** catalogs. Reproduction:

```
GET /api/tenants/demo/catalogs/epistola-demo/templates    → 13 items
GET /api/tenants/demo/catalogs/municipality-demo/templates → identical 13 items
```

Closes #466.

## Root cause

Two layers, one bug — present in two parallel endpoints:

1. **REST controllers** (`EpistolaTemplateApi.listTemplates`, `EpistolaStencilApi.listStencils`) received `catalogId` and built the query without forwarding it.
2. **Queries** (`ListDocumentTemplates`, `ListStencilSummaries`) had no `catalogKey` field, so the SQL `WHERE` only filtered by `tenant_key`.

Sibling endpoints `listThemes` / `listFonts` / `listCodeLists` already pass `catalogKey = CatalogKey.of(catalogId)` to queries that accept it; this PR brings templates and stencils up to that same shape.

## What changed

- **`ListDocumentTemplates`** / **`ListStencilSummaries`**: added optional `catalogKey: CatalogKey? = null` field; SQL conditionally appends `AND <table>.catalog_key = :catalogKey` and binds it. Mirrors the existing `ListThemes` pattern.
- **`EpistolaTemplateApi.listTemplates`** / **`EpistolaStencilApi.listStencils`**: now pass `catalogKey = CatalogKey.of(catalogId)`.
- **Backward compat preserved** for cross-catalog tenant-wide callers (`TenantHandler.kt:55` template count, `LoadTestHandler.kt` x4, `StencilHandler.kt:78`) — `catalogKey` defaults to `null`, no SQL filter, today's behavior unchanged.

## Why not require `catalogKey`?

`TenantHandler` (template count for the tenant overview) and the load-test fixtures legitimately want **all** templates across **all** catalogs for a tenant. Splitting into a separate `CountTenantTemplates` query for one `AND` clause was unnecessary churn; optional parameter matches the precedent in `ListThemes`.

## Cross-surface impact (CLAUDE.md item 11)

- **Web UI** — handler paths under `/templates/*` build composite IDs from path params; unaffected.
- **REST** — this PR. Both `templates` and `stencils` list endpoints now filter correctly.
- **MCP** — does not call these two queries; nothing to wire.

## Tests

New regression coverage:

- `EpistolaTemplateApiIT.list templates is scoped to the requested catalog` — HTTP-level: seeds two catalogs, posts a template into each, asserts each catalog's GET contains only its own slug.
- `ListStencilSummariesCatalogScopeTest` (new) — mediator-level: two catalogs, two stencils; asserts catalog-scoped, cross-catalog, and unfiltered listings each return the right set.

## Test plan

- [x] `./gradlew ktlintCheck unitTest` — green
- [x] `./gradlew :modules:epistola-core:integrationTest :modules:rest-api:integrationTest :apps:epistola:integrationTest` — new tests green; 2 unrelated flakes (`PublishVersionTest`, `ContractVersionCommandsTest$PublishContractVersionTest`) with `Connection is closed` / `EOFException` in `CreateTenant`/`CreateDocumentTemplate`, both pass in isolation
- [x] `pnpm format` — clean
- [ ] Manual smoke against demo:
  ```
  curl -s -H "X-API-Key: $KEY" .../api/tenants/demo/catalogs/epistola-demo/templates | jq '.items | length'
  curl -s -H "X-API-Key: $KEY" .../api/tenants/demo/catalogs/municipality-demo/templates | jq '.items | length'
  ```
  Counts must differ and sum to the previously-returned total.


mvanleijenhorst:
**Request changes** — not blocking-large, but concrete enough to resolve before merge:

1. **Fix `StencilHandler.kt:78` in this PR.** One line (add `catalogKey = catalogFilter`). It's literally the same category of bug as #466, on the JSON branch of the stencil search UI handler. The PR description currently claims this caller is intentionally cross-catalog, but the code reads `catalogFilter` from the query string and then discards it — so the description is misleading. Either fix it here, or update the PR description and open a follow-up issue with a ticket number referenced in the PR body. **Approve** if you address the issue in this PR. **Approve** as well if you argue it out to a follow-up issue with the ticket number in the PR description — just don't leave it unspoken in the current state.

2. **Inline the `test {}` alias** in `ListStencilSummariesCatalogScopeTest.kt:19` — costs nothing, removes one line of indirection that's used exactly once.

sdegroot:
Both points addressed in 843c439b:

1. **Stencil JSON search now passes `catalogKey`.** `StencilHandler.search` at the JSON branch was parsing `?catalog=` and then calling `ListStencilSummaries` without it. The HTMX branch already passed `catalogKey`, so the editor's inline stencil picker (`searchStencils` in `templates/editor.html`, `Accept: application/json`) was the visible regression. One-line fix; added `StencilHandlerHtmxTest.JSON search honours the catalog query param` as deterministic handler-level coverage (per the testing-rules preference for handler tests over browser tests on server-contract assertions). You were right that the PR description misframed this caller as "intentionally cross-catalog" — it isn't, the code shape just didn't read that way. The wider CHANGELOG entry now calls out the JSON-handler branch explicitly.

2. **Inlined the `test {}` alias.** Replaced with `fun ... (): Unit = withMediator { ... }`; JUnit accepts the method because the return type is now explicitly `Unit`.

Both regression tests green locally:
- `StencilHandlerHtmxTest > JSON search honours the catalog query param() PASSED`
- `ListStencilSummariesCatalogScopeTest > list filters by catalogKey when set() PASSED`

PR #472, opened 2026-05-22
feat!: RFC 9457 problem details (type-only) across REST API + UI

whit3st:
## Description

Adopt **RFC 9457** `application/problem+json` (which obsoletes RFC 7807) for error responses across **both** the REST API and the UI, aligned with `epistola-contract` `0.7.0`.

**Contract alignment (`epistola-contract` 0.7.0):**
- The problem **`type` URI is the single machine-readable discriminator** — the separate `code` member was **dropped** (it duplicated `type`, whose slug *is* the code). The contract maps `ProblemDetail`/`ValidationProblemDetail` onto Spring's native `org.springframework.http.ProblemDetail` (no generated DTO); field-level errors use the typed `ValidationError`.

**REST API (`/api/**`):**
- Central problem-type registry (`ApiProblemTypes`), shared builder, framework-level pre-dispatch handling, and API auth (401/403) failures that preserve the Bearer challenge when applicable.
- Public HTML problem documentation at `/errors/{slug}`.

**UI surface (everything except `/api/**`)** — a single `application/problem+json` envelope for structured (fetch/XHR/HMX) callers; HTML error pages for browser navigations:
- `UiExceptionFilter` for handler exceptions.
- UI Spring Security `exceptionHandling` for **401 / 403 / CSRF** (previously the whitelabel `/error` page).
- New `UiHandlerExceptionResolver` for **pre-dispatch** framework errors (404/405/415/406) and **uploads over the limit** (413).
- Editor draft-save now returns `ValidationProblemDetail`; a missing/blank `templateModel` is a `400` problem (was a `500`/NPE).

**Editor (TypeScript):** switches on the problem `type` URI instead of a `code` constant (`save-service`, `binding-errors`, `editor.html`); HMX `responseError` handlers read `detail` (with an `error` fallback).

**Docs:** ADR 0004 amended (RFC 9457 + the corrected Jackson-3 reasoning).

## Related Issue(s)

- RFC 9457 contract alignment in `epistola-contract` (released as `0.7.0`).
- Follow-up: server-side JSONata binding validation gap — epistola-suite#497.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Component(s) Affected

- [x] Documentation
- [ ] CI/CD

## Checklist

## Additional Notes

- **Intentional API breaking change:** error responses negotiate as `application/problem+json` with the **`type` URI as the discriminator** — there is no `code` member.
- Scope is **both** the REST API and the UI surface (the editor's draft-save included).
- Targets `epistola-contract` `0.7.0` (released); branch is merged up to date with `main`.


mvanleijenhorst:
Custom Map<String, Any?> vs. Spring's ProblemDetail

The PR builds error responses as a hand-rolled Map<String, Any?> instead of Spring's built-in org.springframework.http.ProblemDetail. 

Since Spring 6, ResponseEntityExceptionHandler already maps the framework exceptions — 405 (HttpRequestMethodNotSupportedException), 415 (HttpMediaTypeNotSupportedException), 406 (HttpMediaTypeNotAcceptableException), 400/MethodArgumentNotValidException,
404/NoResourceFoundException, type mismatches, unreadable bodies, etc. — to RFC 7807 ProblemDetail out of the box. 

The custom approach 
gives full control over the top-level code extension and key ordering, but it costs ~930 lines of boilerplate and forces the duplication 
between ApiExceptionHandler and ApiFrameworkExceptionResolver (the same five framework exceptions are handled in both places). It's a 
legitimate trade-off, but exactly the kind of decision that belongs in an ADR. So what is the reason we what to use the custom approach vs the spring's default?

---
  Current approach (what's in the PR)
  
  A registry plus a map-builder, then ~50 advice methods that each return ResponseEntity<Map<String, Any?>>:

```
  // ProblemDetails.kt — hand-built body
  fun problemBody(
      request: HttpServletRequest,
      type: ApiProblemType,
      detail: String,
      extensions: Map<String, Any?> = emptyMap(),
  ): Map<String, Any?> {
      val body = linkedMapOf<String, Any?>(
          "type" to type.type.toString(),
          "title" to type.title,
          "status" to type.status.value(),
          "detail" to detail,
          "instance" to request.requestUriWithQuery(),
          "code" to type.code,
      )
      body.putAll(extensions)
      return body
  }
  
  // ApiExceptionHandler.kt — one of ~50 handlers
 @ExceptionHandler(TenantNotFoundException::class)
  fun handleTenantNotFound(ex: TenantNotFoundException, request: HttpServletRequest): ResponseEntity<Map<String, Any?>> =
      problemResponse(request, ApiProblemTypes.TENANT_NOT_FOUND, ex.message ?: "Tenant not found", mapOf("tenantId" to ex.tenantId.value))
  
  // ApiExceptionHandler.kt — 405 handled here...
 @ExceptionHandler(HttpRequestMethodNotSupportedException::class)
  fun handleMethodNotSupported(ex: HttpRequestMethodNotSupportedException, request: HttpServletRequest): ResponseEntity<Map<String, Any?>> {
  /* ... */ }
  
  // ...AND again in ApiFrameworkExceptionResolver.kt for the pre-dispatch case (the duplication)
  private fun handleMethodNotSupported(request, response, ex): ModelAndView { /* same body, different plumbing */ }
  
  The reason ApiFrameworkExceptionResolver exists at all is that @RestControllerAdvice(basePackages = …) only fires once a controller method
  has been matched — so framework exceptions thrown before dispatch (no-handler, wrong method, unsupported media type) escape it. That's a
  real gap, but Spring already solves it.
```

  ---
  Spring-native approach

  The idea: extend ResponseEntityExceptionHandler (it owns all the framework exceptions and already returns ProblemDetail), and centralise the
   Epistola-specific bits — the type URI, the code extension, the instance — in the single handleExceptionInternal hook. That deletes both the
   framework-exception handlers and the separate resolver.


```
  @RestControllerAdvice(basePackages = ["app.epistola.suite.api.v1"])
  class ApiExceptionHandler : ResponseEntityExceptionHandler() {

      // 1. Framework exceptions (405/415/406/400/404/type-mismatch/unreadable-body/…)
      //    are already handled by the superclass and already produce a ProblemDetail.
      //    We only enrich every outgoing body in ONE place:
      override fun handleExceptionInternal(
          ex: Exception,
          body: Any?,
          headers: HttpHeaders,
          statusCode: HttpStatusCode, 
          request: WebRequest,
      ): ResponseEntity<Any>? {
          if (body is ProblemDetail) {
              val type = ApiProblemTypes.forStatus(statusCode) // map 405 -> METHOD_NOT_ALLOWED, etc.
              body.type = type.type
              body.title = type.title
              body.setProperty("code", type.code)
              (request as? ServletWebRequest)?.request?.let { body.instance = it.problemInstance() }
          }
          return super.handleExceptionInternal(ex, body, headers, statusCode, request)
      }

      // 2. Domain exceptions: one small handler that turns an ApiProblemType into a ProblemDetail.
      @ExceptionHandler(TenantNotFoundException::class)
      fun handleTenantNotFound(ex: TenantNotFoundException, request: HttpServletRequest): ProblemDetail =
          problemDetail(request, ApiProblemTypes.TENANT_NOT_FOUND, ex.message ?: "Tenant not found",
              mapOf("tenantId" to ex.tenantId.value))
  }

  // Build a ProblemDetail instead of a Map. Extensions become top-level members
  // automatically via Spring's ProblemDetailJacksonMixin (@JsonAnyGetter on getProperties()).
  fun problemDetail(
      request: HttpServletRequest,
      type: ApiProblemType,
      detail: String,
      extensions: Map<String, Any?> = emptyMap(),
  ): ProblemDetail = ProblemDetail.forStatusAndDetail(type.status, detail).apply {
      this.type = type.type
      this.title = type.title 
      this.instance = request.problemInstance()
      setProperty("code", type.code)
      extensions.forEach { (k, v) -> setProperty(k, v) }
  }
```

  You can go one step further for the ~50 domain exceptions: have each one implement Spring's ErrorResponse (or throw ErrorResponseException)
  so it carries its own ProblemDetail. Then the giant switchboard of @ExceptionHandler methods collapses to near-zero — the exception
  describes itself and Spring renders it:
  
```
  class TenantNotFoundException(val tenantId: TenantKey) :
      ErrorResponseException(
          HttpStatus.NOT_FOUND,
          problemDetail(ApiProblemTypes.TENANT_NOT_FOUND, "Tenant '$tenantId' not found")
              .apply { setProperty("tenantId", tenantId.value) },
          null,
      )
  // No @ExceptionHandler needed — ResponseEntityExceptionHandler renders ErrorResponseException directly.
```

  What this buys you, mapped to my two findings:
  
 
| Concern | Custom Map (current PR) | ResponseEntityExceptionHandler + ProblemDetail |
| :--- | :--- | :--- |
| **Framework exceptions** (405/415/406/400/404) | Handled twice (advice + resolver) | Handled once by the superclass |
| **ApiFrameworkExceptionResolver** | Needed (pre-dispatch gap) | Mostly unnecessary¹ |
| **code / type / instance enrichment** | Repeated per handler | One `handleExceptionInternal` override |
| **Domain exceptions** | One `@ExceptionHandler` each (~50) | `ErrorResponse`/`ErrorResponseException`, near-zero handlers |
| **Content negotiation, etc.** | Hand-wired | Built in |
  
  What you give up / caveats:
  - Key ordering isn't guaranteed in the same way (the LinkedHashMap order). RFC 7807 clients shouldn't depend on key order anyway, so this is
   cosmetic.
  - Manual serialization paths (the Spring Security AuthenticationEntryPoint / AccessDeniedHandler, and the ClientIdentityFilter) sit outside
  MVC, so they still serialize a ProblemDetail by hand via the ObjectMapper. That path depends on the ProblemDetailJacksonMixin being
  registered on the bean to flatten properties to top-level — which is the one thing worth verifying with an integration test .
  - ¹ NoHandlerFoundException only reaches the advice if throwExceptionIfNoHandlerFound is enabled (default in Boot); NoResourceFoundException
   is already handled by the superclass. So the resolver can usually be dropped entirely for /api/**.

mvanleijenhorst:
Minor: ClientIdentityFilter takes a defaulted ObjectMapper = ObjectMapper() parameter; prefer injecting the Spring-managed bean
  (SecurityConfig already passes it, so the default is only a fallback).

`UpdateVersion`, `ArchiveVersion`, and `PublishToEnvironment` gain `template_key` (and
`catalog_key`) in their `WHERE` clauses to prevent cross-template collisions when variant + version
ids collide, with a new `VersionScopingTest`. That is a genuine **correctness / behavioral fix**,
not an error-format change. Bundled into the RFC 7807 PR it can't be reviewed, bisected, or
reverted independently.
**Recommendation:** split it into its own PR/commit and link a dedicated issue.

### `/api/ping` client-identity requirement was relaxed
`ClientIdentityFilter` no longer requires the client-identity headers on `/api/ping` (they are now
mandatory only on `/generation/collect`). This is a behavioral change unrelated to the error
format. It is tested, but not clearly called out.
**Recommendation:** at minimum, mention it in the PR description and the CHANGELOG.

## Tests
### Assert the 403 path as `application/problem+json`
The 403 path (the `accessDeniedHandler` in `SecurityConfig`) is not asserted as
`application/problem+json` end-to-end. The 401 entry point is unit-tested; the 403 path is not.
**Recommendation:** add a targeted assertion that a forbidden `/api/**` request returns
`application/problem+json` with a top-level `code` of `ACCESS_DENIED`.
### Cover the `instance` value without a query string
Only the with-query-string branch of `instance` is currently tested. Add a case asserting that
`instance` equals the request URI when there is no query string, to exercise the other branch of
`requestUriWithQuery()`.

whit3st:
#### Comment 1 — Use Spring's `ProblemDetail`

**Done (Option B).** We now build responses via `ProblemDetail` then convert to `Map<String, Any?>` for serialization. Extending `ResponseEntityExceptionHandler` was blocked by Jackson 3 incompatibility (`ProblemDetailJacksonMixin` uses Jackson 2) and Spring 7.0.6 `@ExceptionHandler` ambiguity. Documented in ADR 0004.

#### Comment 2a — `ClientIdentityFilter` default `ObjectMapper`

**Fixed.** Removed default from both `ClientIdentityFilter` and `ApiKeyAuthenticationFilter`.

#### Comment 2b — Extract query scoping fix

**Fixed.** Extracted into a separate commit (`fix(core): scope template/stencil version queries to owning resource`) ahead of the RFC 7807 commit.

#### Comment 2c — `/api/ping` client-identity relaxation

**Acknowledged.** Mentioned in CHANGELOG. No code change.

#### Comment 2d — Missing 403 `accessDeniedHandler` test

**Fixed.** Added end-to-end test via `ForbiddenTestController` in `CollectEndpointSmokeIT`.

#### Comment 2e — Missing `instance` test (no query string)

**Fixed.** Added in `ProblemDetailsTest`.

PR #485, opened 2026-06-01
fix: prevent nested shell on cancel links

### LINKED ISSUE #483: [Bug]: Cancelling the new API key form renders a duplicate menu bar and footer
### Description

On the **New API key** page (`/tenants/{tenantId}/api-keys/new`), clicking **Cancel** renders the API-keys list page with a duplicate navigation menu bar (and footer) nested inside the original shell. The page becomes visually broken — two menu bars stacked vertically, two footers, and the list content appears inside the `<main>` of the page you were just on instead of replacing it.

### Steps to Reproduce

1. Sign in and open a tenant.
2. Navigate to **API Keys**.
3. Click **New API key**.
4. On the new-key form, click **Cancel** (do not submit the form).
5. Observe the resulting page.

### Expected Behavior

Clicking Cancel returns to the API keys list as a clean, normal navigation — exactly one menu bar, one footer, one `<main>` content area.

### Actual Behavior

The API keys list is rendered **inside** the existing shell's `<main>`, so the page contains:

- The original top-level navigation bar (from the shell of the new-key page).
- A second, duplicate navigation bar (the one belonging to the list page that was just fetched).
- The list page content.
- A duplicate footer.
- Below that, the original footer.

### Component

Frontend (Thymeleaf/HTMX) — `apps/epistola/src/main/resources/templates/api-keys/new.html`.

### Root Cause

Three things combine:

1. The application shell sets `hx-boost="true"` on `<body>` (`apps/epistola/src/main/resources/templates/layout/shell.html:11`), so every link/form does an AJAX swap by default.
2. In `apps/epistola/src/main/resources/templates/api-keys/new.html:13`, the `<form>` declares `hx-target="#api-key-form-area"` and `hx-swap="outerHTML"` (correct for the form's own POST — it swaps the form area for the created-panel fragment on success).
3. The **Cancel** anchor lives **inside that `<form>`** (`new.html:37`), and `hx-target` / `hx-swap` are inherited HTMX attributes.

When Cancel is clicked, HTMX fires a boosted `GET /tenants/{id}/api-keys`. `ApiKeyHandler.list` always returns the full page (shell + nav + `<main>` + footer). Because of the inherited target, HTMX swaps the **whole returned body** (nav + main + footer) into the small `#api-key-form-area` div — which is itself already nested inside the existing shell's `<main>`. The original menu bar stays; the response's menu bar gets injected inside it.

(The **Done** link on the created-panel works correctly because it lives outside any form and therefore inherits no target — it falls back to the boost default of `body` / `innerHTML`.)

### Suggested Fix

Override the inherited target on the Cancel link to match the boost defaults:

```html
<a th:href=\"@{/tenants/{tenantId}/api-keys(tenantId=${tenantId})}\"
   hx-target=\"body\" hx-swap=\"innerHTML\"
   class=\"ep-btn ep-btn-outline\">Cancel</a>
```

### Related — same latent bug in other forms

The same Cancel-inside-form-with-hx-target pattern exists in two other create forms and will reproduce the same defect if the corresponding GET handlers are exercised via Cancel:

- `apps/epistola/src/main/resources/templates/code-lists/new.html:130` (form target `#code-list-form-area`, swap `outerHTML`).
- `apps/epistola/src/main/resources/templates/loadtest/new.html:63` (form target `#form-error`, swap `innerHTML`).

Worth fixing in the same change.

### Environment

- Backend: Spring Boot 4.0 + Kotlin 2.3 (JDK 25)
- Frontend: Thymeleaf + HTMX 2.0.10
- Reproduces in any modern browser; no browser-specific behavior involved.

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported
- [x] I have read the Contributing Guidelines

whit3st:
## Description

Fixes the duplicate/"nested shell" rendering when users click **Cancel** on the create forms for API keys, code lists, and load tests.

### Root cause

The app shell enables `hx-boost` globally on `<body>`, so every link is a boosted AJAX navigation. HTMX's `hx-target`/`hx-swap` are **inherited** attributes, so a `Cancel` `<a>` nested inside a form that scopes its swap to a fragment (e.g. `<form hx-target="#form-area" hx-swap="outerHTML">`) inherited that target. Because a boosted request renders the full `layout/shell` page, Cancel swapped the **entire shell** into the small form `<div>` — rendering a nested nav bar and footer.

### Approach

The first iteration overrode each `Cancel` link with `hx-target="body" hx-swap="innerHTML"`. That works, but it's a per-link patch — applied to only the forms that happen to have a Cancel link today, and easy to forget on the next control added.

This PR instead fixes the bug **class** centrally. A boosted navigation in this app *always* means "replace `<body>` with a full page" (already encoded on the request side in `HtmxResponseBuilder`/`HtmxRender`). A new `HxBoostRetargetFilter` completes that invariant on the response side: any request with `HX-Boosted: true` gets `HX-Retarget: body` + `HX-Reswap: innerHTML`, overriding any inherited form-level target. Links inside `hx-target` forms therefore need **no** special attributes, so the three per-link overrides are removed.

This is safe for inline-error forms: `HX-Boosted` is sent only for boost-driven plain `<a>`/`<form>` navigation. Explicit `hx-post`/`hx-get` requests (the validation-error fragments, cascading dropdowns) are **not** boosted — that distinction is exactly what the existing fragment-vs-full-page logic relies on — so their responses are never retargeted.

It mirrors how the app already handles cross-cutting HTMX concerns (e.g. global CSRF-token injection), rather than per-form markup.

## Related Issue(s)

Fixes #483

## Type of Change

- [ ] Documentation update

## Component(s) Affected

- [x] Frontend (Thymeleaf templates / HTMX)
- [x] Documentation
- [ ] CI/CD

## Checklist

## Additional Notes

**Added**

- `apps/epistola/src/main/kotlin/app/epistola/suite/config/HxBoostRetargetFilter.kt` — central boosted-nav retarget filter.
- `apps/epistola/src/test/kotlin/app/epistola/suite/config/HxBoostRetargetFilterTest.kt` — unit tests: boosted → `HX-Retarget: body`/`HX-Reswap: innerHTML`; non-boosted HTMX and plain requests → untouched.

**Changed**

- Removed the per-link `hx-target="body"` overrides from `api-keys/new.html`, `code-lists/new.html`, `loadtest/new.html`.
- `docs/htmx.md` — new "Boosted Navigation Always Targets `body`" section.
- `.claude/skills/htmx-form/SKILL.md` — note that Cancel links inside `hx-target` forms need no override.
- `CHANGELOG.md` — `Fixed` entry under `[Unreleased]`.


whit3st:
I found this:
 https://htmx.org/attributes/hx-disinherit/ 

Would this work here?

PR #490, opened 2026-06-02
fix: decode submitted form fields as UTF-8

### LINKED ISSUE #484: Scope PopupLoginFilter via FilterRegistrationBean (avoid /* servlet auto-registration)
## Background

`PopupLoginFilter` (`apps/epistola/src/main/kotlin/app/epistola/suite/config/PopupLoginFilter.kt`) exists to persist a `popup=true` marker into the session during the OAuth2 / form‑login redirect chain, so `PopupAwareAuthenticationSuccessHandler` can read it back after the IdP round‑trip. It only cares about two paths:

- `/login`
- `/oauth2/authorization/*`

It is declared as a `@Component` extending `OncePerRequestFilter`, with `@Order(Ordered.HIGHEST_PRECEDENCE)` and no `FilterRegistrationBean`. As a result, Spring Boot's `ServletContextInitializerBeans` auto‑registers it against the servlet container with the default URL pattern `/*` — so it is **invoked on every request**, and only no‑ops for non‑login paths via an internal `if` check inside the method body.

### Why this matters

This combination recently caused a real, user‑visible bug: form‑submitted diacritics in template names (and any other accented form field) were corrupted (`Café` → `CafÃ©`). The filter at `HIGHEST_PRECEDENCE` called `request.getParameter(...)` unconditionally, which forced Tomcat to parse the form body **before** Spring's `CharacterEncodingFilter` had set the request encoding to UTF‑8 — freezing the decode to Tomcat's ISO‑8859‑1 default for every form POST in the app. Once Tomcat has parsed the parameters, later `setCharacterEncoding("UTF-8")` calls are silently ignored.

The immediate symptom has been patched on branch `fix/template-name-diacritics` by guarding the `getParameter` call behind the path check, plus a regression test (`DocumentTemplateRoutesTest.MetadataUpdateEncodingTest`) that round‑trips a diacritic template name through both the form‑create and JSON‑PATCH paths. That fixes the symptom, but the structural fragility remains: any future contributor adding a similar high‑precedence filter that touches parameters would reintroduce the same class of bug.

## Proposal

Register `PopupLoginFilter` explicitly via `FilterRegistrationBean` and restrict its URL patterns to the paths it actually cares about, so it is physically not in the filter chain for unrelated requests:

```kotlin
@Bean
fun popupLoginFilterRegistration(filter: PopupLoginFilter) =
    FilterRegistrationBean(filter).apply {
        urlPatterns = listOf("/login", "/oauth2/authorization/*")
        order = Ordered.HIGHEST_PRECEDENCE
    }
```

Drop `@Component` from `PopupLoginFilter` (otherwise Boot still auto‑registers it for `/*` and you end up with two registrations — one scoped, one global), or construct it inside the `@Bean` method.

## Notes / considerations

- `OncePerRequestFilter` and `@Order` do **not** scope a filter to specific requests — `OncePerRequestFilter` only guarantees "at most once per dispatch," and `@Order` controls relative ordering. The servlet URL pattern is the only mechanism that scopes which requests reach a filter.
- The internal path guard added in the immediate fix can stay as defence‑in‑depth or be removed once the registration scopes the patterns — preference call.
- `MediatorFilter` and `SecurityFilter` follow the same `@Component` + `OncePerRequestFilter` pattern but genuinely need to run on every request (mediator + security context setup), so they should stay as‑is.
- Worth adding a tiny test that asserts `PopupLoginFilter` is **not** invoked on a non‑login path (e.g. via a counter or by inspecting the resolved filter chain), to lock this in against regressions.

## Acceptance criteria

- [ ] `PopupLoginFilter` is registered only for `/login` and `/oauth2/authorization/*` (no global `/*` registration).
- [ ] Existing popup‑login behaviour is preserved (session attribute set in exactly the same cases as today).
- [ ] `MetadataUpdateEncodingTest` and the broader `DocumentTemplateRoutesTest` stay green.
- [ ] (Optional) A test asserting the filter does not fire on non‑login paths.

mvanleijenhorst:
## Problem

Saving a template whose form fields carried diacritics — e.g. a name like `Café` — persisted ISO-8859-1-decoded mojibake (`CafÃ©`). Reading worked fine (stored bytes round-trip back through the JSONB/Jackson UTF-8 path), so the corruption only happened on **save** via form submission.

## Root cause

`PopupLoginFilter` runs at `@Order(Ordered.HIGHEST_PRECEDENCE)` — the same precedence as Spring's character-encoding filter — and called `request.getParameter(...)` on **every** request.

A servlet `application/x-www-form-urlencoded` body is parsed lazily, on the first `getParameter()` call, using whatever `request.getCharacterEncoding()` is at that moment. If it's still unset, Tomcat falls back to the spec default **ISO-8859-1**, and once parsed the encoding is locked — a later `setCharacterEncoding("UTF-8")` is a no-op. Because `PopupLoginFilter` won the same-precedence tie and triggered the parse early, every form POST got decoded as Latin-1 before the UTF-8 setting took effect.

## Fix

Gate the `getParameter()` call behind the `/login` and `/oauth2/authorization/*` path check it already required. The popup flag only ever arrives as a **query** parameter on those entry points, so ordinary form POSTs no longer trigger an early, wrongly-encoded body parse. Login/OAuth behaviour is unchanged.

This single change fixes **all** form-based handlers (templates, themes, fonts, environments, …) since they share the `ServerRequest.form()` path. The JSON editor save path was already UTF-8-safe and is untouched.

## Testing

New `TemplateDraftEncodingHtmxTest` (`apps/epistola`) replicates the exact wire shape a browser sends and asserts diacritics survive the round trip:

- **form POST** of a template `name` (UTF-8 percent-encoded, bare `application/x-www-form-urlencoded`) — fails on `main`, passes with this fix.
- **JSON PUT** of the editor draft model (raw UTF-8 bytes, bare `application/json`) — guards the already-correct path.

```
./gradlew :apps:epistola:test --tests "app.epistola.suite.handlers.TemplateDraftEncodingHtmxTest"
```

Existing `DocumentTemplateRoutesTest` still passes; `ktlintCheck` clean.

---

Relates to #484 (the structural filter-hygiene follow-up — registering PopupLoginFilter via FilterRegistrationBean scoped to /login + /oauth2/authorization/* — stays tracked there; this PR removes the underlying encoding-bug class at the container level instead).


mvanleijenhorst:
### Reviewer note — scope of this fix (trade-off worth knowing)

This is a **surgical** fix: it gates the early `getParameter()` call in `PopupLoginFilter`. Two residual weaknesses to be aware of:

1. **The landmine isn't fully removed.** Any *future* filter added at `@Order(Ordered.HIGHEST_PRECEDENCE)` that reads a request parameter would re-introduce the identical bug — the first `getParameter()` parse locks Tomcat to ISO-8859-1 before the encoding filter (same precedence) sets UTF-8.

2. **Regression coverage is path-specific.** `TemplateDraftEncodingHtmxTest` exercises the template form/JSON paths, but would *not* catch a new filter breaking the theme/font/environment forms, even though they share the same vulnerable `ServerRequest.form()` path.

**Durable alternative (deliberately out of scope here):** give the app's custom filters `HIGHEST_PRECEDENCE + N` so Spring's `OrderedCharacterEncodingFilter` is guaranteed to run first and force UTF-8 before any app code can trigger a body parse — neutralising the whole class rather than this one call site.

Tracked as a follow-up: see linked issue. Reasonable to merge this as the immediate fix and harden separately.

mvanleijenhorst:
### Follow-up: landmine removed (commit d65e802f)

Rather than just track the residual weakness, I removed the whole class of bug.

**`RequestEncodingConfig`** pins the servlet container's default request/response encoding to UTF-8 at the `ServletContext` level — established *before any filter or servlet runs*. Since a form body is parsed lazily on the first `getParameter()` using whatever encoding is set at that moment, fixing the default up front means the first parse (in any filter, at any `@Order`) decodes as UTF-8. The filter-ordering dependency is gone.

The earlier `PopupLoginFilter` gate and Spring's `CharacterEncodingFilter` stay as complementary layers (the gate also avoids eager body parsing on every request).

**Proof it's durable:** `RequestEncodingLandmineTest` installs a hostile filter at `@Order(HIGHEST_PRECEDENCE)` that reads a parameter on every request — the exact condition that caused the original bug — and asserts a diacritic form POST still round-trips. I also verified manually by reverting the `PopupLoginFilter` gate (recreating the real landmine): the form test still passed with the container-level fix in place, where before it reliably produced `CafÃ©`.

Full `unitTest + integrationTest` suite passes with the new context-level encoding bean in place.

Issue #506, opened 2026-06-07
Dedicated hub OTLP metrics registry + per-tenant metrics toggle


sdegroot:
## Context

Part of the fleet-monitoring work. The transport-agnostic **foundation is already merged** on `feat/metrics`:

- Common identity tags on every metric (`service`/`instance`/`installation_id`/`environment`/`version`) via `MetricsConfig` + a shared `NodeIdentity`.
- Percentile histograms for latency timers.
- Actuator isolated on a separate management port in prod (`/livez`/`/readyz` probes on the main port).

This issue tracks the **future hub-push layer** (deferred from the foundation PR). Decision recap from planning:

- **Transport: OTLP push.** Remote instances behind NAT can't be reliably scraped.
- **Hub leg is completely separate (Option A):** a dedicated in-app `OtlpMeterRegistry` bean we own end-to-end — its own `OtlpConfig` (hub URL, `ek_*` auth from `epistola-support`, DELTA temporality, tuned `step`/`batchSize`), its own `MeterFilter`s, its own resource attributes. Isolated from anything the customer does.
- **Customer leg is theirs:** make the container bring-your-own-agent friendly (export OTel resource-attribute defaults so an attached OTel Java agent inherits `service.instance.id=nodeId` etc.; default the autoconfigured SDK OTLP to hub-only so a customer agent owns the customer leg with no double export). Customers can also scrape `/actuator/prometheus`.

## Scope

- [ ] Dedicated hub `OtlpMeterRegistry` bean gated by `epistola.metrics.export.hub.{enabled,url,...}` (default off), DELTA temporality, `ek_*` auth reusing the `epistola-support` hub credentials.
- [ ] Per-tenant `metrics` feature toggle: add `METRICS = FeatureKey.of("metrics")` to `KnownFeatures` + default in `FeatureDefaults`/`application.yaml` (same pattern as `feedback`).
- [ ] Hub-only `MeterFilter` on the hub registry that denies meters whose `tenant` tag belongs to a tenant with the `metrics` feature disabled (inputs from `FeatureToggleService`, cached). The customer leg is not gated.
- [ ] OTel resource-attribute defaults so an attached OTel Java agent inherits a consistent identity; default the autoconfigured SDK OTLP to hub-only when an agent is present (no double export).
- [ ] `docs/` bring-your-own-agent guide (expected `OTEL_*` env, hub-only default, recent-agent note for virtual threads).
[...]

PR #507, opened 2026-06-07
feat(metrics): fleet-monitoring foundation for multiple remote instances

sdegroot:
## Why

We will run multiple Epistola instances remotely and need to monitor them well, eventually forwarding metrics to **epistola-hub**. This PR lands the transport-agnostic foundation plus the operational coverage needed to make fleet monitoring trustworthy. The hub-push layer itself is deferred to #506.

## What

**P0 — foundation**
- **Common identity tags on every metric** (`service`, `instance`, `installation_id`, `environment`, `version`) via a `MetricsConfig` `MeterFilter`, so series from N instances reporting into one backend are distinguishable (and survive OTLP push as resource/datapoint attributes). Node identity is resolved by a single app-wide `NodeIdentity` in core — `GetServerInfoHandler` and the support hub client now share it.
- **Percentile histograms** for the key latency timers + `http.server.requests`, so p95/p99 aggregate correctly across the fleet.
- **Separate management port in prod** (`management.server.port: 4040`): actuator (health/info/prometheus) off the public app port, kept cluster-internal. Kubernetes probes move to `/livez` / `/readyz` on the **main** port (`add-additional-paths`, always-on) so they never depend on the management context.
- **Node-id removed from the Helm chart**: a Deployment value is identical for every replica and can't provide per-pod identity — it's resolved at runtime from the pod hostname instead.

**P1 — operational coverage**
- **Health indicators**: `contentStore` reachability + `jobPoller` heartbeat, contributing to `/actuator/health` for alerting (deliberately not the readiness group — jobs fail over via `SKIP LOCKED` and the UI/API still serve).
- **Scheduled-cron metrics**: `ScheduledTaskMetrics` wraps partition maintenance, stale-job recovery, and the stale-consumer reaper (`epistola.scheduled.task.duration{task,outcome}`) — a cron that silently stops firing shows as the absence of recent samples. Feedback sync gains `epistola.feedback.sync.attempts{outcome}`.
- **Installation-wide gauges**: `epistola.installation.{tenants,templates,themes,catalogs,stencils,fonts,environments}`, published once per round by the Postgres-advisory-lock holder (non-holders report `NaN`, which Micrometer omits); a `MeterFilter` strips `instance` so the series is stable across leader changes.

## Deferred
- Dedicated hub OTLP registry + per-tenant `metrics` toggle + bring-your-own-agent friendliness → #506.

## Testing
- New: `NodeIdentityTest`, `MetricsConfigTest`, `JobPollerHealthIndicatorTest`, `ScheduledTaskMetricsTest`, `InstallationStatsPublisherIT`; extended `PrometheusEndpointTest` (common tags present, `/livez`+`/readyz`, both health indicators wire up).
- Full `./gradlew unitTest integrationTest` green; `ktlintCheck` clean; `helm template` renders (probes on `/livez`/`/readyz`, management port 4040 not in the Service).

## Notes
- `management.server.port` only applies under the `prod` profile; `add-additional-paths` is always-on so probe paths work in every profile.
- `observability.prometheus.port` Helm default is now `4040` (set to `4000` if running without the `prod` profile).


apps/epistola/src/main/kotlin/app/epistola/suite/config/MetricsConfig.kt
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * `installation_id` is read from `app_metadata`, so this bean must not be
+     * created before Flyway has run — hence the explicit `@DependsOn`
+     * (JDBI has no implicit depends-on like `JdbcOperations` does).
+     */
+    @Bean
+    @DependsOn("flywayInitializer")
+    fun commonMetricTags(
+        nodeIdentity: NodeIdentity,
+        installations: InstallationService,
```
sdegroot, line 62: What is this installation service? We typically do everything lke this with commands/queries right? Why a service class here? I guess this was created earlier. If there is no good reason, we should create a separate ticket to resolve this deviation.

apps/epistola/src/main/kotlin/app/epistola/suite/config/MetricsConfig.kt
```diff
@@ -0,0 +1,102 @@
...
+    internal fun commonTagsFilter(
+        nodeId: String,
+        installationId: String,
+        environment: String,
+        version: String,
+    ): MeterFilter {
+        val tags = listOf(
+            Tag.of("service", "epistola-suite"),
+            Tag.of("instance", nodeId),
+            Tag.of("installation_id", installationId),
```
sdegroot, line 89: Theoretically installation_id may change at runtime. Resulting into metrics that have a stale installation_id

apps/epistola/src/main/kotlin/app/epistola/suite/config/SecurityConfig.kt
```diff
@@ -154,7 +161,11 @@ class SecurityConfig(
         http
             .authorizeHttpRequests { authorize ->
                 authorize
-                    // Public endpoints
+                    // Public endpoints. /livez + /readyz are the Kubernetes probe
+                    // paths add-additional-paths exposes on the MAIN port (the
+                    // actuator endpoints themselves move to the management port in
+                    // production); /actuator/** entries cover single-port profiles.
+                    .requestMatchers("/livez", "/readyz").permitAll()
                     .requestMatchers("/actuator/health/**", "/actuator/info").permitAll()
```
sdegroot: this requesrtMatcher, is that still relevant?

modules/epistola-core/src/main/kotlin/app/epistola/suite/observability/ScheduledTaskMetrics.kt
```diff
@@ -0,0 +1,41 @@
...
+ * Records execution count, duration, and outcome for background scheduled tasks
+ * (maintenance crons, cleanups, sync loops).
+ *
+ * Without this a cron that silently stops firing — or starts failing every run
+ * — is invisible across a remote fleet. Emits the timer
+ * `epistola.scheduled.task.duration` tagged `task` and `outcome`
+ * (`success` | `failure`). Alert on the **absence** of recent samples to catch
+ * a stalled scheduler, and on the failure rate to catch a broken one.
+ */
+@Component
```
sdegroot: why is this a spring component? why not an extension method on meterRegistry?

apps/epistola/src/main/kotlin/app/epistola/suite/config/MetricsConfig.kt
sdegroot: Good catch — this is a pre-existing deviation (no `GetInstallation` query exists; `InstallationService` is the only read path, predating this PR). Out of scope for the metrics work, so I filed #509 to migrate it to a CQRS query. Keeping `InstallationService` here for now.

apps/epistola/src/main/kotlin/app/epistola/suite/config/MetricsConfig.kt
sdegroot: It's invariant by design: `InstallationService`'s contract is that the `app_metadata` row is Flyway-seeded once and fixed for the lifetime of the database (every pod reads the same value). And Micrometer common tags are static by nature — mutating `installation_id` mid-process would split one series into two, which is worse. A genuine identity change (DB swapped under a live pod) is picked up on the next pod restart. Expanded the comment to spell this out (5513cfff).

apps/epistola/src/main/kotlin/app/epistola/suite/config/SecurityConfig.kt
sdegroot: It's redundant — removed in 5513cfff. The order-0 `managementSecurityFilterChain` (`EndpointRequest.toAnyEndpoint()`, permit-all) handles all actuator endpoints in single-port profiles, and in prod they're on the separate management port, so they never reach this UI chain. Kept `/livez`+`/readyz` since those are root-level add-additional-paths probes not matched by `toAnyEndpoint()`.

modules/epistola-core/src/main/kotlin/app/epistola/suite/observability/ScheduledTaskMetrics.kt
sdegroot: Agreed — converted to a `MeterRegistry.recordScheduledTask(task) {}` extension function in 46077089 (no bean; the schedulers inject `MeterRegistry` directly).

PR #534, opened 2026-06-10
feat: list bullet styles, nesting, and reliable marker rendering (#401)

mvanleijenhorst:
## Summary

Closes the remaining work on #401 (list styling parity between the `datalist` component and lists inside the `text` component), and fixes two concrete defects found along the way.

### Bullet styles + reliable marker rendering
- The `datalist` gains a **`bulletStyle`** prop (`disc`/`circle`/`square`/`dash`), mirroring the Text component's `bullet_list` `listStyle`. Shown in the inspector only when `listType: bullet` (new editor-only `InspectorField.visibleWhen`), and surfaced to MCP via the regenerated `component-registry.json` + a new `styled-bullet-list` example.
- **Markers now render in the list's content font**, not iText's WinAnsi default (Helvetica), which lacks `circle` (U+25CB) / `square` (U+25A0) — they were silently dropping from the PDF while the editor showed them. When the resolved content font itself lacks the glyph (e.g. no font family configured → Helvetica), the marker **falls back to the bundled Liberation Sans** (`FontCache.fontCoveringOrFallback`). So all four styles render regardless of the configured font.

### Nested lists
- A `bullet_list`/`ordered_list` nested inside a `list_item` ("een opsomming in een opsomming") was dropped by `ProseMirrorConverter.convertListItem` (paragraph-only). It now recurses at arbitrary depth.
- The editor had **no gesture to nest a list item**. Added a list keymap — **Tab** sink, **Shift-Tab** lift, **Enter** split (prosemirror-schema-list) — plus matching **indent/outdent buttons** in the bubble menu (shown only inside a list).

### Demo
- Simple Letter: Attachments became a data-driven `datalist` (`bulletStyle: square`); added a nested "Key points" list with a `circle` sublevel. Demo catalog `5.7.1` → `5.8.0`, fingerprint regenerated.

## Test plan
- [x] `./gradlew build` (all modules, unit + integration + UI) — green
- [x] `ListStyleParityTest` — parity for all four bullet styles + decimal; **all glyphs (incl. circle/square) render with no font configured** (fallback)
- [x] `NestedListRenderTest` — nested bullet/ordered lists render their inner items
- [x] `list-keymap.test.ts` — Tab/Shift-Tab/Enter on the custom `bullet_list` schema
- [x] `BundledCatalogFingerprintTest` / `CatalogIntegrationTest` — demo `5.8.0`
- [x] Editor vitest (1777) + `pnpm format:check` + ktlint — green
[...]

PR #542, opened 2026-06-11
refactor: deterministic cluster scheduling (trigger/engine split) + event-driven job drain

sdegroot:
## Problem

The native scheduler introduced in #537 broke the test suite's concurrency model (red builds like #534):

- **Installation-wide state vs tenant-scoped isolation.** Integration tests run test classes concurrently and rely on everything being `tenant_key`-scoped. The `cluster_timers` / `cluster_tasks_scheduled` tables and several jobs (partition maintenance, stale-job recovery, stats) are installation-wide — a background poller firing mid-test races concurrently-running test classes on state that tenancy cannot isolate.
- **The wall-clock trigger is structurally incompatible with test time.** The test clock is `ScopedValue`-bound to the test thread (`EpistolaClockExtension`); the three background `@Scheduled` loops (heartbeat 2s, timer poll 1s, task poll 1s) run on scheduler threads and evaluate `next_due_at <= now` against the **system clock**. No interval tuning fixes that — every scheduler IT was already inflating poll intervals to 60s/1h and driving the engines by hand.
- The migration also deleted the deterministic job-drain trigger: `GenerateDocument` inserts a `PENDING` row and nothing nudges the `JobPoller` except the autonomous loop (`GenerationRequestCreatedEvent` existed with no publisher; `epistola.generation.synchronous` was a dead key).

## Solution: separate the trigger substrate from the engines

Scheduling is *when* to invoke; the engines own *what* runs. The engines (`ClusterNodeHeartbeatScheduler`, `ClusterTimerScheduler`, `ClusterScheduledTaskScheduler`) lose their `@Scheduled` annotations and become plain, directly-invokable components. A `ClusterSchedulingDriver` owns the cadence, selected by `epistola.cluster.scheduling-substrate`:

- **`wall-clock`** (production default): `WallClockClusterSchedulingDriver` ticks the engines on the same `epistola.cluster.*-interval-ms` delays as before — production behavior is unchanged.
- **`test`**: `DeterministicClusterScheduling` (in `modules/testing`, wired by `IntegrationTestBase`) replaces the loop with explicit, synchronous virtual-time execution on the test thread:

```kotlin
// "a day passes and due scheduled work runs"
scheduling.advanceTimeBy(Duration.ofHours(25))
scheduling.runDue()            // run what is already due
testClock.advanceBy(...)       // move time without firing anything
```

`runDue()` loops timers + tasks to quiescence (engines' `poll()` now returns the claimed count). Selection follows the repo's property-gated `@ConditionalOnMissingBean` idiom so the two drivers are mutually exclusive regardless of configuration ordering. Scheduler ITs drop their poll-interval-inflation workarounds.

## Event-driven job drain (prod improvement too)

`GenerateDocument` / `GenerateDocumentBatch` publish `GenerationRequestCreatedEvent` / `GenerationBatchCreatedEvent` after the insert commits; the new `GenerationRequestDrainListener` nudges `JobPoller.requestDrain()` immediately. Jobs start right away in production (instead of up to `interval-ms` later) and deterministically in tests. The recurring `core.document-job-poller` task remains the safety-net fallback.

## Found along the way: 39 silently-skipped tests

A Kotlin expression-body `@Test` returning a value (ending in an AssertJ chain) is only a WARNING-severity discovery issue for JUnit — it logs and **never runs the test**. 36 such methods in `epistola-core` and 3 in `apps/epistola` (`ApiKeyAuthenticationFilterTest`) had been silently skipped and rotted:

- `CancelGenerationJob` tests asserted the old `false`-return contract (now throws domain exceptions)
- `GenerateDocument(Batch)` tests expected `IllegalArgumentException` (now domain not-found exceptions)
- `JobPollerIntegrationTest` couldn't even boot (no `@SpringBootTest(classes=...)`) and its ad-hoc principal lacked a `users` row
- Awaitility `until`-lambdas queried the mediator on the polling thread without a bound principal
- the asset reverse-chronological test tied on the frozen test clock

All resurrected and repaired, and `junit.platform.discovery.issue.severity.critical=warning` now fails discovery on any future instance.

## Test plan

- [x] Cluster scheduler/timer/registry/partition ITs green driving engines directly (no loop)
- [x] New virtual-time IT: `scheduling.advanceTimeBy(...)` runs a task that became due
- [x] `DocumentGenerationIntegrationTest` / `JobPollerIntegrationTest` green — generation completes via the create-event → `requestDrain` path with no wall-clock driver in context
- [x] `JobPollerIntegrationTest` (outside `IntegrationTestBase`) required-injects `WallClockClusterSchedulingDriver`, guarding the production default substrate
- [x] Full `./gradlew unitTest integrationTest` across all modules
- [x] `ktlintCheck` + `pnpm format:check`

## Also: raised the integration-test JVM heap 512m → 1g

While verifying the branch, `:apps:epistola:integrationTest` OOMed (`GC overhead limit exceeded`) **intermittently**, with a different victim test each run — and `main` did the same on a 10-core machine (1 of 2 runs). Not introduced here: JUnit's class-level concurrency boots distinct Spring contexts in parallel (one per unique `@SpringBootTest` config), and each context load runs a full-classpath component scan whose classfile metadata transiently costs ~100MB+. On the 2-core CI runner ≤2 contexts load at once and 512m holds; on a 10-core dev box ~7 can, blowing the heap. Bumped `integrationTest` (and the catch-all `test`) worker heap to 1g in `epistola-kotlin-conventions.gradle.kts`. CI behavior is unchanged; local runs stop flaking.

Relates to #534, #537.


modules/epistola-core/src/test/kotlin/app/epistola/suite/documents/JobPollerIntegrationTest.kt
```diff
@@ -80,13 +88,14 @@ class JobPollerIntegrationTest {
     )
 
     private fun <T> withAuthentication(block: () -> T): T = MediatorContext.runWithMediator(mediator) {
+        TestPrincipalUsers.ensure(mediator, testUser)
```
sdegroot: why do we have TestPrincipalUsers and SecurityContext both passing the testUser?

modules/testing/src/main/kotlin/app/epistola/suite/testing/DeterministicClusterScheduling.kt
```diff
@@ -0,0 +1,88 @@
...
+ * [IntegrationTestBase]) so it is mutually exclusive with the production
+ * wall-clock driver regardless of configuration-class ordering; a test can opt
+ * back into the real loop with
+ * `@TestPropertySource(properties = ["epistola.cluster.scheduling-substrate=wall-clock"])`.
+ */
+@TestConfiguration(proxyBeanMethods = false)
+class DeterministicSchedulingTestConfiguration {
+
+    @Bean
+    @ConditionalOnProperty(
```
sdegroot, line 79: Do we need this conditional? Should we not always enable it in test?

modules/epistola-core/src/test/kotlin/app/epistola/suite/documents/JobPollerIntegrationTest.kt
sdegroot: Good catch — they did two different things (`ensure` materialises the `users` row for the audit FKs, `SecurityContext.runWithPrincipal` binds the principal for authz), but that pair is exactly what `TestPrincipalUsers.runWithPrincipal` bundles. Collapsed to that single helper in 50b343da. (It is hand-rolled here only because this test deliberately does not extend `IntegrationTestBase` — it skips `FakeExecutorTestConfiguration` to exercise real PDF generation — so it cannot use the base `runAs`/`withAuthentication`.)

modules/testing/src/main/kotlin/app/epistola/suite/testing/DeterministicClusterScheduling.kt
sdegroot: It is always enabled in test — `IntegrationTestBase` sets `epistola.cluster.scheduling-substrate=test`, so every test extending it gets the deterministic driver. The reason it is a property gate (rather than an unconditional `@Bean` that relies on the wall-clock driver stepping aside via `@ConditionalOnMissingBean`) is robustness: `@ConditionalOnMissingBean` alone is order-dependent between an `@Import`ed test config and the component-scanned production `@Configuration`, so it can register both drivers or neither depending on configuration ordering — the same race we already hit and documented in `FeedbackSyncFallbackConfiguration`. The substrate property makes the two mutually exclusive regardless of ordering: in test exactly the deterministic driver registers and the wall-clock one is gated off. It also leaves a clean escape hatch for the one test that wants the real loop (`JobPollerIntegrationTest`, which does not extend the base). Happy to drop the property and rely on bean-presence if you would rather accept that ordering caveat — but I would keep it.

PR #544, opened 2026-06-11
feat(catalog): whole-catalog EF-style wire-format migration framework + import error surfacing

### LINKED ISSUE #493: Ensure catalog functionality works 100%
- catalog schema version should be included
- backwards compatibility (older versions of epistola catalogs should be able to be imported in newer versions!)

mvanleijenhorst:
The **catalog-wide, EF-Core-migrations-style framework** for the catalog exchange wire format. On import, an ordered chain upgrades a whole catalog from the `schemaVersion` it was exported at to the version this instance understands — applied on the **JSON tree before binding**, so an export produced by an older instance is mechanically upgraded without the producer being reachable.

This branch is the **framework + wiring + operator polish**, shipping with an **empty chain** (no migration content yet); worked example migrations live on the `example/catalog-wire-format-migration` branch.

Architecture: [ADR 0007](docs/adr/0007-catalog-wire-format-migrations.md) · Roadmap: [plan](docs/plans/catalog-wire-format-migrations.md) · Wire contracts: [docs/exchange/](docs/exchange/README.md)

## Catalog-wide version axis

There is **one** catalog-wide wire `schemaVersion` (currently `4`). The **manifest is authoritative** for it, and `CatalogContentBuilder` stamps the **same** number on every resource detail so each file is self-describing — but the whole bundle moves together. There is **no** independent per-resource version. `CATALOG_SCHEMA_VERSION` / `CATALOG_BASELINE_SCHEMA_VERSION` hold the single window. (This reverses an earlier per-part draft — see ADR 0007.)

## What's in this PR

**1. Framework** (`app.epistola.suite.catalog.migrations`)
- `CatalogSchemaMigration` — one step per `from → from+1`, with `migrateManifest` and `migrateResourceDetail` pure `JsonNode` transforms (a step can reshape the manifest and/or any resource-detail tree).
- `CatalogSchemaMigrator` — collects steps into **one chain**, validates it is contiguous + total **at startup** (Flyway-like: a malformed chain fails app boot), and gates the payload's `schemaVersion` once.
- `CatalogSchema{TooNew,TooOld,Unknown}Exception`.
- `CATALOG_SCHEMA_VERSION` / `CATALOG_BASELINE_SCHEMA_VERSION` constants; **uniform export stamping** in `CatalogContentBuilder` (manifest + every detail get the same number).

**2. Both chokepoints wired** — manifest **and resource details** are migrated before binding, gated against the single catalog version, at the ZIP path (`ImportCatalogZipHandler`) and the remote/classpath path (`CatalogClient`, covering install / browse / upgrade-check / fingerprint).

**3. Robustness** — invalid JSON and a detail whose own `resource.type` disagrees with the manifest entry both surface as `CatalogSchemaUnknownException`; a wire-version gate failure on a resource detail rejects the **whole** import (rethrown past the per-resource catch) rather than being downgraded to a single `FAILED` resource.

**4. Operator-facing error surfacing** — a rejected payload (too-new / too-old / unrecognised) is actionable on every surface:
- **REST** `POST /api/tenants/{id}/catalogs/import`: dedicated RFC 9457 problems — `…/errors/catalog-schema-too-new` (`version`, `supportedVersion`), `…/catalog-schema-too-old` (`version`, `baselineVersion`), `…/catalog-schema-unknown` — instead of a generic 400.
- **Web UI** import dialog: the remediation message renders inline (same `alert-error` slot as a stencil conflict).
- **MCP**: read-only, no import tool → no change.

## Gate (one catalog version)

| `schemaVersion` | Behaviour |
|---|---|
| `== current` | bind directly |
| `baseline ≤ v < current` | run the chain, then bind |
| `v < current`, empty chain | pass through unchanged (transitional) |
| `> current` | reject — `TooNew` (upgrade this instance) |
| `< baseline` | reject — `TooOld` (re-export from a current source) |
| not valid JSON, or missing / non-integer | reject — `Unknown` |

Migration never touches `release.fingerprint` / `release.version` (`schemaVersion` is excluded from the fingerprint by construction).

## Docs

- **[`docs/version-axes.md`](docs/version-axes.md)** — a map of the easily-conflated "version" concepts: the editorial **template / contract / stencil** publish versions (instance-local), the **catalog release version + fingerprint** (content release, for upgrade detection), and the **catalog-wide wire `schemaVersion`** (the exchange format shape). The **stencil version** is the one bridge that travels on the wire as data.
- A **wire-format version-gate** section in [`exchange/README.md`](docs/exchange/README.md#wire-format-version-gate) (the catalog-wide gate table + operator surfaces) and a fingerprint-transparency note in [`catalog-versioning.md`](docs/catalog-versioning.md#fingerprint-algorithm).

## Not in this PR

- **Migration content** — the chain is empty here. Worked examples (a `v3 → v4` field rename and a `v4 → v5` no-op version bump) live on the `example/catalog-wire-format-migration` branch.
- **"Upgraded from wire vN" success note** — deferred follow-up.

## Testing

- `CatalogSchemaMigratorChainTest` / `…GateTest` — chain integrity + the single catalog-version gate.
- `CatalogSchemaProblemMappingTest` — REST problem mapping.
- `StencilVersionImportConflictTest` — incl. a too-new resource detail aborting the whole import.
- Catalog import/export integration tests green; `ktlintCheck` + `oxfmt --check` clean. Up to date with `main`.


sdegroot:
I wonder: would it be sensible to have the migration in the contract-repo itself maybe?

mvanleijenhorst:
> I wonder: would it be sensible to have the migration in the contract-repo itself maybe?

It make sense there is already a issue created for it #572

modules/epistola-core/src/main/kotlin/app/epistola/suite/catalog/CatalogConstants.kt
```diff
@@ -1,12 +1,25 @@
...
+ * [CATALOG_SCHEMA_VERSION] (manifest + every resource detail move together). See
+ * `docs/adr/0007-catalog-wire-format-migrations.md` and `docs/exchange/`.
+ *
+ * `CATALOG_BASELINE_SCHEMA_VERSION == CATALOG_SCHEMA_VERSION` today — no
+ * migrations exist yet, so the chain is empty and every payload binds as-is.
  */
-const val CATALOG_MANIFEST_SCHEMA_VERSION: Int = 4
+
+/** The wire schema version this instance emits, and upgrades older imports to. */
+const val CATALOG_SCHEMA_VERSION: Int = 4
```
sdegroot, line 19: im guessing we should actually move this to the contract repo. Better to do that in a separate PR though

modules/epistola-core/src/test/kotlin/app/epistola/suite/catalog/CommittedCatalogSchemaWindowTest.kt
```diff
@@ -0,0 +1,69 @@
...
+ * *below* baseline imports today only because of the empty-chain leniency in
+ * [app.epistola.suite.catalog.migrations.CatalogSchemaMigrator] (sub-current +
+ * empty chain → pass through); the moment a migration chain exists the same
+ * fixture would be rejected as `TooOld`. This test makes that latent trap a build
+ * failure: our own committed content never relies on the leniency crutch — only
+ * genuinely-old *external* payloads do.
+ */
+class CommittedCatalogSchemaWindowTest {
+
+    private val mapper = jsonMapper { addModule(kotlinModule()) }
```
sdegroot, line 26: should we use our default json mapper maybe? to avoid potential differences in how jackson is configured

sdegroot: i guess we don't do that in other tests either. Something to be fixed later

sdegroot: Agreed, leaving as-is. Worth noting for this specific test it's a non-issue regardless: it only does `readTree(...)` + `isIntegralNumber`/`asInt`, and `kotlinModule()` only affects *typed data-class binding*, not tree parsing — so this test is immune to Jackson-config drift. The real cleanup is repo-wide (a shared test-mapper factory for the ~6 core catalog unit tests that copy `jsonMapper { addModule(kotlinModule()) }`), which is its own hygiene PR rather than part of this feature.

modules/epistola-core/src/main/kotlin/app/epistola/suite/catalog/CatalogConstants.kt
sdegroot: new issue created #596

PR #564, opened 2026-06-13
feat(security): encrypt stored credentials at rest with key rotation

### LINKED ISSUE #397: Encrypt at-rest credentials for subscribed catalogs and URL-sourced code lists
## Context

Two columns currently store auth credentials (Bearer tokens, API keys) in plaintext:

- `catalogs.source_auth_credential` (V4 migration — pre-existing) — credential for refreshing a subscribed remote catalog.
- `code_lists.credential` (V29 migration — new in #392) — credential for refreshing a URL-sourced code list.

Both follow the same pattern: `auth_type IN ('NONE', 'API_KEY', 'BEARER')` plus a `credential TEXT` column. The new code-list column inherited the catalog precedent. @mvanleijenhorst flagged this on the code-list column in #392 review — the concern applies equally to the catalog column.

## What to build

Encrypt both columns at rest:

1. Pick an envelope-encryption shape — likely AES-GCM with a per-row IV, KEK either env-injected or AWS KMS / GCP KMS depending on deploy target. Match whatever the rest of the platform plans for at-rest secrets (this is the first one we'd introduce, so the choice sets precedent).
2. Migration that re-encrypts existing rows. Pre-prod (per `CLAUDE.md`) so a one-shot rewrite is acceptable.
3. Application-layer encrypt-on-write / decrypt-on-read for both columns. Likely an attribute converter or a JDBI argument factory + column mapper pair so the encryption is invisible at the command/query layer.
4. Tests confirming the columns contain ciphertext at rest (raw SQL read returns non-plaintext) and round-trip decryption returns the original value.

## Considerations

- Both columns are nullable (`auth_type = NONE` rows store `NULL`); the encryption layer needs to pass NULLs through.
- Key rotation: design the format so an old-IV / old-KEK row can be re-encrypted lazily on next read, or batch-rotated.
- Backup/restore: encrypted columns mean the KEK must be backed up separately or be deterministic from a stable secret.

## Why both columns in one issue

Doing only `code_lists.credential` would leave the existing `catalogs.source_auth_credential` gap intact. The fix shape is the same for both — better to address them together with one consistent pattern than to introduce a divergence.

sdegroot:
Closes #397.

## What

Encrypts the sensitive credentials Epistola stores so a database dump alone never exposes a usable secret:

- **Catalog credentials** (`catalogs.source_auth_credential`)
- **Code-list credentials** (`code_lists.credential`)
- **Hub credentials** (support tier's `app_metadata` entry)

> **API keys are intentionally *not* encrypted** — they're a verify-only secret, so they stay **hashed** (SHA-256, plaintext never stored). Encrypting them would be strictly weaker. They gain only a performance cache (below).

## How

- **New `modules/epistola-crypto`** (dependency-light, JDBI-free): `CredentialCipher` does AES-256-GCM with a fresh 96-bit nonce per encryption, a 128-bit tag, and the key id bound as AAD. Stored as a self-describing envelope that fits the existing `TEXT` columns:
  ```
  enc:v1:<keyId>:<base64url(nonce ‖ ciphertext ‖ tag)>
  ```
- **Transparent persistence:** domain types carry a redacted `Secret` field; JDBI `SecretArgumentFactory`/`SecretColumnMapper` (registered in `JdbiConfig`) encrypt on write / decrypt on read. `AppMetadataService.setEncrypted`/`getEncryptedAs` cover the hub-credentials JSONB entry.
- **Key rotation (no downtime):** the keyset (`epistola.encryption.keys` + `primary-key-id`) holds every key for decryption while new writes use the primary. Rotate = add key → promote → `ReencryptCredentials` backfill (idempotent `SystemInternal` command, also upgrades legacy plaintext) → `VerifyNoStaleKeyIds` → retire old key.
- **API key cache:** short-TTL bounded `ApiKeyAuthCache` (default 60s/10k, caches negative misses, re-checks `isUsable()` on hit, invalidated on revoke) keeps the hot `/api/**` path fast.

## Config & deployment

- **Local dev:** a fixed throwaway key in `application-local.yaml` so encrypted data survives restarts.
- **Other non-prod profiles with no keys:** ephemeral in-memory key + warning.
- **`prod` profile:** fails fast at startup if enabled with no key.
- **Helm:** keyset wired from a Kubernetes Secret as indexed `EPISTOLA_ENCRYPTION_*` env on the app Deployment only (migration Job never touches ciphertext). `encryption.enabled=true` by default. Note: this is the **encryption keyset**, not application-generated API keys (those are runtime-minted per tenant and stored hashed — nothing to inject).

See [`docs/encryption.md`](docs/encryption.md).

## Tests

- 14 crypto unit tests (round-trip, envelope format, nonce uniqueness, rotation, tamper/AAD-swap detection, plaintext pass-through, key validation, prod fail-fast).
- Integration tests proving **ciphertext-in-DB + transparent decrypt** (code-list credential) and encrypted-metadata round-trip.
- Existing catalog/code-list/API-key filter tests still green; architecture-enforcement tests pass for the new commands/queries.
- `ktlintCheck`, `pnpm format:check`, and `helm lint` all clean.


mvanleijenhorst (approved):
Looks good

PR #571, opened 2026-06-14
refactor(ui): consolidate page headers into a shared fragment

whit3st:
## Description

Consolidates page-header markup into a single shared Thymeleaf fragment so list, create, and detail pages render consistent header chrome instead of each template hand-rolling its own `<div class="page-header">`. The shared chrome lives in the **`epistola-web`** toolkit module so feature modules can use it without depending on the host app.

**What changed:**

- **Shared `epistola-web/page-header` fragment** — one header fragment with explicit, optional arguments: required `pageTitle`, plus optional `pageDescription`, search box (`searchPlaceholder` / `searchUrl` / `searchTargetId`), `actions`, and a back link (`backLinkHref` / `backLinkLabel`). Title always renders with `data-testid="page-title"` so tests stay stable across migrated pages.
- **Chrome lives in `epistola-web`, not the host.** The `page-header` fragment moved out of `apps/epistola` into the `epistola-web` toolkit, so the support feature modules that consume it no longer pick up a template-level dependency on the host app (modules → toolkit, never modules → host). The search box stays in the host — it's a host list-page feature the support modules don't use.
- **Shared `icon` fragment moved too.** The generic `icon` primitive — used unconditionally by host and feature-module templates alike (back-links, action buttons) — also moved into `epistola-web`, and its reference path was renamed `fragments/icon` → `epistola-web/icon` across all usages.
- **Each page owns its actions.** Per-page action buttons are defined as a local `th:fragment` in the page's own template and passed to the header via `actions=~{::name}`. Feature-module actions (backups, feedback) live in those modules' own templates rather than a central host registry, so there's no host→module knowledge and no module→host fragment dependency.
- **~40 templates migrated** across the host app and the support feature modules (backups, feedback, upgrading, support overview) to call the shared fragment.
- **Responsive header** — a long description no longer collides with the action buttons: a gap separates the two columns, the content column wraps instead of overflowing, the action buttons keep their size (and wrap when there are several), and below `56rem` (where the nav also goes mobile) the header stacks into a single column.
- **CSS** — header spacing is owned by `.page-header` in `design-system/components.css` (`align-items: start`, vertical padding, `.content h1/p` styling); the now-redundant `.main-content h1` margin was removed from `shell.css`.
- **Button consistency** — the code-list detail actions were aligned on `ep-btn` (they carried legacy `btn btn-outline` / `btn btn-destructive`).
- **Back-links replace breadcrumbs.** Every breadcrumb in the app was only one level deep, so it duplicated the page title and amounted to a back-link already; standardising on the back-link makes navigation consistent across pages. The breadcrumb markup and its CSS were removed.

**Design boundary held intentionally:** the fragment owns only generic page chrome. Page-specific state — read-only notes, metadata rows, status badges, source URLs — is kept *below* the header rather than forced into the fragment, since those patterns don't repeat uniformly enough to share yet.

## Visual changes (not a pure no-op)

Mostly structural, but two deliberate visual changes ship with it:

- `.page-description` previously had no CSS, so page descriptions were unstyled. They now render as muted small text (`text-sm`) via `.page-header .content p`.
- The entitlements section description on the support overview was given the same muted small styling for consistency.
- The header now stacks (title/description above, actions below) on viewports narrower than `56rem`.

## Deferred to a follow-up (inline-styling cleanup)

Two review suggestions are intentionally **not** done here. Both are inline-style cleanups that belong to a separate, dedicated PR that will sweep inline styles across the app, so doing them one-off here would be out of scope:

- **The `<p>` at `support/overview.html:51`** isn't actually the *page's* description — it describes the **Entitlements** section specifically (it even points at the adjacent *Refresh* button), so migrating it into the page header would be wrong. For now it's given inline muted small styling so it matches the other descriptions; giving it a proper semantic class is part of the inline-styling cleanup.
- **The inline-styled Entitlements section header in `overview.html`** (`<div class="ep-row" style="…">`) keeps its inline styles for now; extracting a reusable `.section-header` class will land with that same inline-styling PR rather than be a one-off here.

## Related Issue(s)

Relates to #492

## Type of Change

- [ ] Documentation update
- [x] Refactoring (structural; with the minor intentional visual/responsive changes noted above)

## Component(s) Affected

- [ ] Backend (Spring Boot/Kotlin)
- [x] Frontend (Thymeleaf templates + CSS)
- [ ] Documentation
- [ ] CI/CD

## Checklist

## Additional Notes

- Aside from the muted-description styling and the narrow-viewport stacking noted above, page headers render the same.
- Action buttons are defined locally per page (and per feature module) and passed into the header as a fragment expression — there is no shared central registry of action fragments.


mvanleijenhorst (changes requested):
**Overall:** Strong, worthwhile consolidation. The `page-header` fragment is well designed — explicit optional params, `actions` passed as a fragment expression, and a stable `data-testid` so tests survive the migration. The breadcrumb→back-link change is justified (every breadcrumb was one level deep, so no hierarchy is lost) and the hover detail is a nice touch. Two architectural items to address before merge, plus a few small cleanups.

## Architecture

### 1. Move `page-header` (and `page-actions`) out of the host into `epistola-web`

`page-header.html` currently lives in `apps/epistola/.../fragments/`. The moment the support modules consume it, they pick up an implicit template-level dependency on the **host app** — exactly the coupling `epistola-web` (the shared UI toolkit both the host and feature modules already depend on) exists to absorb. Please move the shared chrome fragment to `epistola-web` so the dependency direction is clean (modules → toolkit, never modules → host).

### 2. Keep page-specific actions with the page, not in a central host registry

`page-actions.html` (187 lines) currently defines **feature-module-specific** action fragments — `backup-list-actions`, `feedback-list-actions`, `feedback-detail-actions` — inside the host, which the support modules then reference back. That inverts the documented boundary (*feature modules ship their own UI; the host composes*): the host ends up carrying module routes/permissions (`/tenants/{id}/backups`, `/feedback/submit-form`, `isManager`, `status`), and the module template depends on a host fragment at the same time.

The header already supports the clean alternative — `actions=~{...}` takes any fragment expression — so each page can define its actions locally:

```html
<!-- backups/list.html (in the backups module itself) -->
<div th:replace="~{epistola-web/page-header :: page-header(
        pageTitle='Backups',
        actions=~{::backup-actions})}"></div>

<th:block th:fragment="backup-actions">
    <form th:action="@{/tenants/{tenantId}/backups(tenantId=${tenantId})}" method="post">
        <button type="submit" class="ep-btn ep-btn-primary" th:disabled="${status != 'OK'}"> … </button>
    </form>
</th:block>
```

Concretely:

- **Feature-module actions** (`backup-*`, `feedback-*`) move back into their own module templates as local fragments. This removes the host→module knowledge and the module→host fragment dependency in one go.
- **Host-only actions** (api-keys, assets, templates, themes, stencils, …) may stay in the host, but prefer inlining them as local `th:fragment`s in each page template too — that lets `page-actions.html` go away entirely rather than persist as a god-file.
- Drop the redundant `template-page-actions(tenantId)` indirection — it only re-`replace`s `new-template`; collapse to one fragment.

### 3. Button class inconsistency (carried over, now centralized)

Most action fragments use `ep-btn ep-btn-*`, but `code-list-detail-actions` uses legacy `btn btn-outline` / `btn btn-destructive`. While relocating these, align them on `ep-btn`.

## UX / cleanup

### 4. "Renders the same" isn't quite accurate

`.page-description` has no CSS definition anywhere — descriptions were previously unstyled (`<p>`). The new `.page-header .content p` rule gives them `text-sm` + muted. That's a small, fine visual *change*, but please reflect it in the PR description rather than calling it a no-op.

### 5. Leftover `page-description` in `support/overview.html:51`

The top description was migrated into the header, but a second `<p class="page-description">` lower in the same file keeps the old unstyled class → now inconsistent with the migrated descriptions. Migrate it or give it a real class.

### 6. Replace the inline-styled section header with a class

In `overview.html`, the secondary `.page-header` (with an `<h2>`) became `<div class="ep-row" style="justify-content:space-between;…">`. Understandable now that `.page-header` is reserved for the H1, but extract a `.section-header` class instead of inline styles so the pattern doesn't get copy-pasted around.

### 7. Visual check on long descriptions

With `align-items: start` and no max-width on `.content`, a long left-column description (e.g. Backups) can run up against the right-aligned action button. Please confirm it holds up on narrower viewports.

## Verdict

Requesting changes on **#1–#2** (the module boundary — header to `epistola-web`, actions local to each page/module). #3 and #4–#7 are small cleanups to fold in along the way. Everything else looks good.

whit3st:
@mvanleijenhorst Thank you for the feedback, I completely forgot about the decoupled modules! My mistake and considering we have decoupled modules, having each action in its own page makes more sense now. I was going to fix button inconsistency in another PR but i can do that here might as well. Taking a look at cleanups too.

whit3st:
@mvanleijenhorst A follow-up question: I know it is out of scope for this PR but fragments/icons are also used in decoupled modules. same problem with what I've done and fixing. Should I also move fragments/icon to epistola-web/icon or that should be done in a separate PR? I could fix that here might as well?

I think my question really is fragments/icon needs to be moved in epistola-web/icon. Should I do that in this PR or in a separate PR 😄

whit3st:
@mvanleijenhorst Addressed the issues and cleanups in the feedback, skipped some of them with a reasoning.  My question about fragments/icon still stands, even thou I did it. I can strip it out for a new PR always. Let me know what do you think! Thank you again

mvanleijenhorst:
Keep it in this PR — don't strip it out.

It's the exact same coupling problem #1 was about: a feature module reaching into a **host** fragment (`fragments/icon`). The whole point of this PR is to make "modules → toolkit, never modules → host" actually *true*. If we move `page-header` but leave `icon` behind, the boundary is only half-fixed — the support modules still depend on a host fragment, so the invariant this PR sets out to establish wouldn't actually hold on merge. I'd rather it land atomically than leave a known violation for "later."

It's also the same *kind* of change that's already in here (move the file, rename the reference path) and mechanically low-risk — `icon` is a generic primitive used unconditionally, so it's a straight find/replace with no behavioral branch. The wide blast radius is just touched files, not added risk.

Contrast that with the inline-styling cleanups (the `overview.html` section header and the semantic description class) — those I *do* want in a separate PR, because they're a genuinely different concern (styling debt swept across the whole app) rather than the module→host decoupling this PR is about. The icon move isn't a different concern; it's the same decoupling, so it belongs with it.

👍 Leave it in.

mvanleijenhorst (approved):

Issue #28, opened 2025-12-26
Create OpenAPI 3.1 specification


sdegroot:
## Summary
Design and document the REST API using OpenAPI 3.1 specification (spec-first approach).

## Details
- Create OpenAPI 3.1 spec for all public API endpoints
- Document request/response schemas
- Include authentication requirements
- Add examples for all endpoints

## Acceptance Criteria
- [ ] OpenAPI 3.1 spec file created
- [ ] All public endpoints documented
- [ ] Request/response schemas defined
- [ ] Examples provided for each endpoint
- [ ] Spec is validated and serves as source of truth

## Technical Notes
- Spec-first approach was chosen over code-first
- SDKs will be generated from this spec

Issue #32, opened 2025-12-26
Implement RFC 7807 Problem Details for errors


sdegroot:
## Summary
Standardize API error responses using RFC 7807 Problem Details format.

## Details
- Return consistent error format for all API errors
- Include type, title, status, detail, and instance fields
- Add extension fields for validation errors

## Acceptance Criteria
- [ ] All API errors return Problem Details format
- [ ] Content-Type: application/problem+json
- [ ] Validation errors include field-level details
- [ ] Error responses documented in OpenAPI spec

## Example Response
```json
{
  "type": "https://epistola.app/errors/validation",
  "title": "Validation Error",
  "status": 400,
  "detail": "The request body contains invalid data",
  "instance": "/api/v1/tenants/1/templates/1",
  "errors": [
    {"field": "dataExamples[0].name", "message": "must not be blank"}
  ]
}
```

whit3st:
This issue is in `post-1.0` milestone but in a umbrella issue which is in `1.0` milestone. @sdegroot can you tell me whether this is pre or post 1.0?

sdegroot:
We should have this in 1.0. Ill update it

Issue #30, opened 2025-12-26
Expose input parameters (JSON Schema) via API


sdegroot:
## Summary
Provide API access to template input parameter definitions (JSON Schema).

## Details
- Expose dataModel (JSON Schema) through the template API
- Allow updating dataModel via API
- Validate dataExamples against schema on update

## Acceptance Criteria
- [ ] dataModel accessible via GET template endpoint
- [ ] dataModel updatable via PATCH endpoint
- [ ] Validation errors returned when examples don't match schema
- [ ] Schema follows JSON Schema 2020-12 standard

## Technical Notes
- dataModel already exists as JSONB column
- Validation already implemented server-side

Issue #29, opened 2025-12-26
Expose template CRUD via REST API


sdegroot:
## Summary
Formalize existing template operations as proper REST endpoints.

## Details
Expose the following operations via REST:
- `GET /api/v1/tenants/{tenantId}/templates` - List templates
- `POST /api/v1/tenants/{tenantId}/templates` - Create template
- `GET /api/v1/tenants/{tenantId}/templates/{id}` - Get template
- `PATCH /api/v1/tenants/{tenantId}/templates/{id}` - Update template
- `DELETE /api/v1/tenants/{tenantId}/templates/{id}` - Delete template

## Acceptance Criteria
- [ ] All CRUD operations available via REST
- [ ] JSON request/response bodies (not HTML)
- [ ] Proper HTTP status codes
- [ ] Tenant isolation enforced

## Technical Notes
- Some endpoints already exist but return HTML
- Need JSON variants for API consumers

PR #577, opened 2026-06-17
feat(backup): faithful per-tenant backup/restore

### LINKED ISSUE #516: Backups capture published catalog content only — consider including drafts
## Limitation

Catalog backups (and the compatibility/Upgrading snapshots) are built by `BuildTenantSnapshot`, which reuses `ExportCatalogZip` per catalog. The catalog export serializes **published** content, so a tenant's **unpublished drafts** (new resources never published, and edits to published resources that haven't been re-published) are **not included** in a snapshot.

Because restore is destructive (`RestoreTenantSnapshot` → `PurgeTenantCatalogs` then re-import), drafts are lost on restore twice over: they aren't captured in the backup *and* the current working copy is wiped and replaced with the snapshot's published content.

This is currently **by design** — the snapshot is byte-for-byte the catalog export so restore feeds it straight back through `ImportCatalogZip`, and the compatibility check intentionally tests published catalogs. It's documented on the Support → Backups page, in the restore confirmation dialog, and in the CHANGELOG.

## Consider

Whether backups should capture unpublished drafts too (full-fidelity backup), which would require either:
- a draft-aware snapshot serialization + a draft-aware restore path (diverging from the export/import primitive, and likely a separate "full" export mode so compatibility keeps testing published content), or
- a different backup mechanism (e.g. a DB-level per-tenant snapshot) instead of the export-based one.

Both are larger than the current export-based design; this issue tracks the decision and the limitation.

## Notes
- `pendingChanges` (`ListCatalogsForManagement`) flags catalogs whose working copy drifted from the last release — a partial signal (it misses never-released draft-only catalogs) that could power a best-effort "you have unpublished changes that won't be backed up" warning if we keep the published-only design.

Split out of the catalog-sync work (PR #512).

### LINKED ISSUE #519: Subscribed catalogs lose their subscription state on snapshot restore
## Problem

Tenant backups/restore (`BuildTenantSnapshot` → `RestoreTenantSnapshot`) build the snapshot by reusing `ExportCatalogZip` per catalog and feed the ZIPs back through `ImportCatalogZip`. The catalog export is a **lossy projection** — for a **SUBSCRIBED** catalog it drops the entire subscription identity, so restore produces a broken "zombie" catalog:

1. **Source connection lost.** On restore, `ImportCatalogZip(catalogType=SUBSCRIBED)` calls `insertSubscribedCatalog(...)`, whose INSERT only lists `id, tenant_key, name, description, type, created_at, updated_at`. `source_url`, `source_auth_type`, `source_auth_credential` land **NULL**. With no `source_url`, `CheckCatalogUpgrade` / `PreviewCatalogUpgrade` / `UpgradeCatalog` all no-op — the subscription can never be upgraded or drift-checked again.

2. **Drift baseline corrupted.** `ExportCatalogZip` stamps `release.version`/`release.fingerprint` from `GetLatestCatalogRelease`, which reads the **AUTHORED-only** `catalog_releases` table. A subscribed catalog has no release rows, so its export is always `0.0.0-dev` + a *locally* recomputed fingerprint (`ExportCatalogZip.kt:72-87`). On restore, `updateSubscribedInstalledState` then overwrites `installed_release_version ← "0.0.0-dev"` and `installed_fingerprint ← <local fp>` — but those columns are meant to hold the **source-side** values recorded at register/upgrade time (e.g. `1.4.2` + the publisher fingerprint). Restore corrupts the upgrade/drift baseline.

### Root cause (shared with #516)

This is the metadata-side counterpart of #516 (backups capture published content only, dropping drafts). Both stem from the same fact: **the snapshot is byte-for-byte the catalog export**, which was designed for catalog *exchange*, not *backup fidelity*. #516 anticipates the remedy — *"draft-aware snapshot serialization … diverging from the export/import primitive."* This issue is the first deliberate divergence: the snapshot must carry subscription state the export structurally cannot.

## Constraints

- **The auth credential must never enter the snapshot.** `source_auth_credential` is a plaintext secret and snapshot archives are uploaded to the external hub. We carry everything *except* the credential.
- **`epistola-contract` must not change.** Subscription identity belongs to the internal `SnapshotCatalogEntry` (epistola-core), not the public `CatalogManifest` exchange format — putting source URL/auth there would leak subscriber state into the publish format. The hub stores the archive as opaque bytes (`catalog_sync.proto` `SnapshotHeader` carries only fingerprint/count/size), so no proto change either.

## Suggested solution

**Make the snapshot authoritative for a subscribed catalog's install-state; install content via the ZIP but strip the import's authority over subscription identity.**

1. **Carry full subscription state in the manifest (minus the credential).** Add to `SnapshotCatalogEntry` (all nullable, populated only for SUBSCRIBED): `sourceUrl`, `sourceAuthType`, `installedReleaseVersion`, `installedFingerprint`, `installedResourceFingerprints`, `installedAt`. Populate in `BuildTenantSnapshot` from the `Catalog` (it already carries all of them). Bump `SNAPSHOT_SCHEMA_VERSION` 1 → 2.

2. **Stop the import from owning subscription state during restore.** Add a restore-only flag to `ImportCatalogZip` (mirroring the existing `validateCrossCatalogDeps`), e.g. `manageSubscribedInstallState = false`: still create the row + install resources, but **skip `updateSubscribedInstalledState`** — the caller owns `source_*`/`installed_*`.

3. **Reapply authoritative state on restore.** New command `RestoreSubscribedCatalogState` doing a single local `UPDATE` (no network call) that sets `source_url`, `source_auth_type`, `installed_*` from the snapshot, leaving `source_auth_credential` NULL. `RestoreTenantSnapshot` calls it after each SUBSCRIBED import; bump `SUPPORTED_SCHEMA_VERSION` 1 → 2. (v1 archives restore unchanged — nullable fields default to null.)

4. **"Reconnect needed" UX for auth-requiring subscriptions.** Add a derived `reconnectNeeded` boolean to `ListCatalogsForManagement` (`type='SUBSCRIBED' AND source_url IS NOT NULL AND source_auth_type<>'NONE' AND source_auth_credential IS NULL`) — credential never leaves the DB, only the boolean. Show a badge in `templates/catalogs/list.html`; the operator re-enters the secret via the existing Subscribe dialog (`RegisterCatalog`'s `ON CONFLICT` upsert rewrites the credential). NONE-auth subscriptions restore fully functional, no action needed.

### Scope / notes

- No DB migration (pre-production). No REST/MCP change (snapshot/restore not exposed there). No `epistola-contract`/proto change.
- Not representable in the demo catalog (backup/restore infrastructure, not a catalog *resource* capability) — covered by integration tests instead.
- Tests: subscribed NONE-auth round-trip preserves real source version (not `0.0.0-dev`); credential string absent from archive bytes (security-critical); auth-required restores credential-less with `reconnectNeeded=true`; v1 archive still restores; reconnect via re-subscribe flips the flag.

### Key files

- `modules/epistola-core/.../catalog/snapshot/TenantSnapshot.kt`
- `modules/epistola-core/.../catalog/snapshot/BuildTenantSnapshot.kt`
- `modules/epistola-core/.../catalog/snapshot/RestoreTenantSnapshot.kt`
- `modules/epistola-core/.../catalog/commands/ImportCatalogZip.kt`
- `modules/epistola-core/.../catalog/commands/RestoreSubscribedCatalogState.kt` *(new)*
- `modules/epistola-core/.../catalog/queries/ListCatalogsForManagement.kt`
- `apps/epistola/.../templates/catalogs/list.html`

Related: #516.

sdegroot:
## Why

Backups today are a thin wrapper over **catalog export**, which is the wrong tool for a backup:

- Export captures only the **latest published** version of each template/contract — no drafts, no history.
- Restore **renumbers versions** (`MAX+1`) — everything comes back as `v1`, breaking any externally-pinned version number (the same hazard `f32f0bc9` fixed for the result sequence).
- Restore **purges and reimports** catalogs, cascade-deleting all generated-document history (`documents` FK `template_versions ON DELETE CASCADE`).

This PR makes backups **faithful and identity-preserving**, distinct from export/import. Export stays a publishing format used only by the Upgrading/compatibility feature.

## What

**Reflective backup primitive** (`epistola-support-backups`, `tenantbackup/`):
- `BuildTenantBackup` / `RestoreTenantBackup` discover tenant-scoped tables, FK insert order, primary keys, and per-column types from `information_schema` at runtime — the dump adapts to schema migrations automatically; only genuine special cases are hand-coded.
- Captures **full version history with exact version numbers**, drafts, env activations, API-key hashes, feature toggles, and asset blobs. An INCLUDE/DENY overlay classifies every tenant-scoped table; a **drift-guard test** fails the build when a new tenant-scoped migration is left unclassified.
- **Restore is a merge, not a wipe**: upsert backup rows + delete-only-absent, so documents/generation pinned to *unchanged* versions **survive** the cascade. Handles the `tenants`↔`themes` cycle, deferred/RESTRICT FKs, partition-child exclusion, and `content_store` asset blobs.
- **Schema-stamp gated** (same-schema undo, not cross-version rollback/DR) and the whole artifact is **encrypted at rest** with the existing `CredentialCipher` keyset.

**Decoupling** (`epistola-support-backups`):
- Backups scheduler + UI repointed to the faithful primitive, stored **locally** in a new `tenant_backups` table behind a `TenantBackupStore` port (a hub transport can replace it later) with `TenantBackupService` (build → fingerprint-dedup → retain N).
- Upgrading keeps the export snapshot (`TenantSnapshotSyncService`) for its hub compatibility checks, so the two features now toggle independently.

## Tests

Testcontainers integration tests (all green): round-trip fidelity + version-number preservation, **cascade-survival** (a document pinned to an unchanged version survives restore), schema-stamp gate refusal, topology drift guard, and the service dedup/restore round-trip. `ktlintCheck`, `pnpm format:check`, and the app's `MediatorWiringTest`/`ApplicationClockUsageTest` pass.

## Open items (not blocking the primitive)

- **Hub transport is deferred** — backups are stored locally (appropriate for the "correct mistakes" purpose; not offsite DR). A hub adapter needs an epistola-hub-side artifact-kind contract.
- **Feature gating unchanged** — `SUPPORT_BACKUPS` stays hub-entitlement-gated as today (conservative). Making local backups available without hub entitlement is a small follow-up in `KnownFeatures`.

See [`docs/tenant-backup.md`](docs/tenant-backup.md).

## Resolves

- Closes #516 — backups now capture full version history including drafts.
- Closes #519 — subscribed catalogs keep their subscription state (incl. encrypted credential) across restore.

---

## Added since opening

- **Hub transport.** A `HubTenantBackupStore` syncs backups to epistola-hub when the support tier is on (the OSS local store is the fallback), via the new hub `TenantBackupService` (hub PR #28). Live-verified end to end on a local hub.
- **Restore across compatible schema versions** (not just the exact same one). A per-migration `schema-backup-compatibility.yaml` (default deny) declares `backward`/`forward` flags; backward reads the live file, forward reads flags the backup snapshots into its manifest (format v2). The structural column check stays as the backstop. The Backups page flags/blocks a backup that can't restore on this build instead of a generic error.
- **Graceful hub-down handling.** The Backups page renders a notice instead of 500-ing when the hub is unreachable / not entitled.
- **Entitlement rename** `support-upgrading` → `support-compatibility-check` (pairs with hub PR #28); a migration re-keys existing `feature_toggles` rows.


mvanleijenhorst (commented):
Reviewed for architecture, security, and tests. Strong, well-layered PR. Recommendations before merge:

1. **Tenant-isolation test (blocking).** Every integration test uses a single tenant. The restore's "DELETE rows absent from the backup" path — especially `deleteAbsent` for the `tenants`/blob/`content_store` paths, which scopes by key prefix rather than `tenant_key` — is untested cross-tenant. Add: backup tenant A, restore A while tenant B holds its own data, assert every B row (catalogs, template_versions, assets, content_store blobs, feature_toggles) is unchanged.

2. **Enforce authentication in `TenantBackupCrypto.unwrap` (Medium security).** A non-`enc:` (plain base64) artifact currently bypasses AEAD entirely — even on an encryption-enabled install — and is trusted on restore (`TenantBackupCrypto.kt:29-33`, via the `CredentialCipher.decrypt` pass-through at `:48`). Assert `CredentialEnvelope.isEnvelope(...)` when the keyset is enabled.

3. **Fix stale coupling/prose.** Drop the unused `api(project(":modules:epistola-support-snapshots"))` dep + its comment in `epistola-support-backups/build.gradle.kts` (no main source imports snapshots anymore — the decoupling is real in code but not in the module graph). Also correct the "core primitive (epistola-core)" references in the PR description, `docs/tenant-backup.md`, and the migration comment in `V20260616222023…sql:7-9` — the code correctly lives in `epistola-support-backups`.

4. **Follow-ups (non-blocking):** retain-N (`pruneToRetention`) test, an encryption tamper-rejection test, a drift-guard positive test (proving it trips on an unclassified table, not just that the current schema is clean), and SQL identifier hardening in `TableRowCodec` (escape `"` in `q()` + allowlist udt names, so restore safety doesn't depend solely on `validateColumns` ordering).

sdegroot:
@mvanleijenhorst thanks — all addressed in d7c6e8e2:

**1. Tenant-isolation test (blocking)** — added `CrossTenantIsolationIntegrationTest`: backs up tenant A, diverges it, restores A while tenant B holds its own catalog, template (→ versions), asset + `content_store` blob, and a feature toggle, then asserts B's rows + blob bytes are unchanged. Confirms both the `tenant_key`-scoped delete and the key-prefix blob delete isolate correctly.

**2. Enforce auth in `unwrap` (Medium)** — `TenantBackupCrypto.unwrap` now refuses a non-`enc:` artifact when the keyset is enabled (exposed `CredentialCipher.enabled`), so a plain-base64 blob can't bypass AEAD. Covered by `TenantBackupCryptoTest` (round-trip / non-envelope rejection / tamper rejection / disabled pass-through). Note this overlaps hub #30 (byte-cipher rewrite of `unwrap`); the guard stands until then.

**3. Stale coupling/prose** — dropped `api(project(":modules:epistola-support-snapshots"))` and now depend on `epistola-support` directly (the module only needed it for the `EpistolaHubClient` bean). Corrected the "epistola-core primitive" references in the migration comment, the `TenantBackupService` KDoc, and the PR description.

**4. Follow-ups (done in this PR too)** — `pruneToRetention` retain-N test; crypto tamper-rejection test; a **positive** drift-guard test (creates an unclassified tenant-scoped table in a rolled-back tx and asserts `resolve` refuses); and SQL identifier hardening in `TableRowCodec` (`q()` doubles embedded quotes; type names are validated against a strict pattern before `::cast` interpolation). While adding the retention test I noticed `save` stamped `captured_at` from the clock at save time — switched it to the artifact's own build-time `capturedAt`.

**CodeQL** — removed the unread `tenantEntry` local in `MergeRestoreTables`.

The remaining Tier-3 perf/cleanup items (restore batching, byte-cipher, archive-codec dedup) are tracked as epistola-hub #29–#32.

PR #603, opened 2026-06-22
feat(design-system,ui): align brand identity with the Epistola website

whit3st:
## Description

Aligns the app's look with the Epistola website on two fronts:

**1. "Peach Paper" color scheme.** Repaints the shared design-system tokens to a warm, paper-like palette — cool gray neutrals become warm taupe/brown and the blue accent becomes terracotta (`#b6684e`). The change is confined to the `--ep-stone-*` and `--ep-terracotta-*` ramps in `tokens.css`; semantic aliases, success/warning/danger colors, and the white panel surface are unchanged, so every surface that routes through the tokens (app **and** editor) warms up at once. The raw palette ramps were renamed to match their new hues (`--ep-blue-*` → `--ep-terracotta-*`, `--ep-gray-*` → `--ep-stone-*`). This is a light-only restyle — Epistola has no dark theme.

**2. Brand mark.** The Epistola logo now ships as `/design-system/epistola-logo.svg` (served via `processResources`) and is used as the navbar brand icon and the browser favicon, replacing the placeholder envelope.

Also in scope:
- Native form controls are tinted with the brand accent via `accent-color`. The WebKit search-clear (×) button ignores `accent-color`, so it's explicitly recolored to the warm neutral (it otherwise read cool against the new palette).
- Small cleanups in touched files: a stale blue fallback in `placeholder.css` (`#2563eb` → `#b6684e`) and the undefined `--ep-bg-input` reference dropped (`→ --ep-white`).

## Related Issue(s)

N/A — standalone brand-alignment pass.

## Type of Change

## Component(s) Affected

- [x] Documentation

## Checklist

## Screenshots (if applicable)

image

image

## **I also changed the editor at the same time. If this is a no-go, let me know please. Also, it would be better if you use the app with this color-palette. Thanks!**

image

## Additional Notes

- **No tests added** — this is a visual restyle. Verified via `oxfmt` format-check, a repo-wide token sweep (zero surviving `--ep-gray-*` / `--ep-blue-*` references; every `var(--ep-*)` in changed CSS resolves), and the existing handler/render tests that exercise nav + shell.
- The token rename is mechanical and routes through unchanged semantic aliases, so component behavior is untouched.
- A few **pre-existing** dangling token references elsewhere (`--ep-bg-secondary` / `--ep-text-secondary` in `error/*.html`, `--ep-radius` in a stencil panel) are out of scope here — they were already undefined on `main`.


apps/epistola/src/main/resources/templates/layout/shell.html
```diff
@@ -4,6 +4,7 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title th:text="${pageTitle} ?: 'Epistola'">Epistola</title>
+    <link rel="icon" type="image/svg+xml" th:href="@{/design-system/epistola-logo.svg}">
```
mvanleijenhorst: The favicon is only added here, but `login.html`, `tenants/list.html`, `templates/editor.html`, and the `error/*.html` pages each have their own `<head>` and don't inherit it — so the **login screen** and the **tenants landing page** (the two most common entry points) show no favicon and fall back to a 404 `/favicon.ico`.

All of those pages already include `~{fragments/styles :: styles}`, so moving this `<link rel="icon">` into **`fragments/styles.html`** covers every page in one place (and the duplicate here can then be dropped).

apps/epistola/src/main/resources/templates/layout/nav.html
```diff
@@ -6,7 +6,7 @@
         <div class="app-nav-bar">
             <div class="app-nav-left">
                 <a th:href="@{/tenants/{id}(id=${tenantId})}" class="app-nav-logo" title="Tenant home">
-                    <th:block th:replace="~{epistola-web/icon :: icon(name='mail', class='ep-icon')}" />
+                    <img class="ep-icon ep-icon-lg" th:src="@{/design-system/epistola-logo.svg}" alt="Epistola" />
```
mvanleijenhorst, line 9: `alt="Epistola"` duplicates the adjacent visible `<span>Epistola</span>`, so a screen reader announces "Epistola Epistola". The wordmark already conveys the name, so the logo is decorative here — use `alt=""`.

apps/epistola/src/main/resources/templates/layout/shell.html
whit3st: Yes, I also realized this. Fix is coming!

apps/epistola/src/main/resources/templates/layout/nav.html
whit3st: `aria-label` should be the better option here I think. An image tag with empty alt is never good 😄 . `aria-label="Epistola"` will bypass anchor and image tag descriptions when anchor tag focus by a screen reader. Thank you for the heads up tho! I missed this as well.

sdegroot:
Looking good! Very on brand. I am a bit worried about readability though. There's not much contrast between buttons, icons, text, etc. anymore. Also the live-preview (center part) should probably be white if the background of the document is white.

whit3st:
TL;DR: A bit darker version of previous color palette.

Few before afters

### Editor:

image


image

### Tenants:

image


image

### Templates list:

image

image

mvanleijenhorst:
think it looks great. But the colors are almost the same as that of error. For color blind it is hard. But for not color blind it is also difficult.

image

Reproduction step:
Go to templates and create a new template. Fill in a slug with MAX length.
And you will get a simular error message

sdegroot:
> I think it looks great. But the colors are almost the same as that of error. For color blind it is hard. But for not color blind it is also difficult.
> 
> image

I'll try something

PR #611, opened 2026-06-23
feat(stencils): upgrade published templates to a newer stencil, with filter/pagination

### LINKED ISSUE #598: [Bug]: upgrade stencil seems broken (v1 remains cannot bulk update)
### Description

We should be able to upgrade multiple templates to the latest stencils 

### Steps to Reproduce

publish two templates with the same v1 stencil 
change in one of the templates the v1 to v2 stencil
go to stencil overview - choose the stencil 
upgrade all - it won't even allow you to upgrade

### Expected Behavior

We should be able to upgrade multiple templates to the latest stencils 

### Actual Behavior

We cannot select templates that use v1. This may have something to do that those templates are not draft ?

### Component

Backend (Spring Boot/Kotlin)

### Environment

_No response_

### Logs / Screenshots

_No response_

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported
- [x] I have read the [Contributing Guidelines](../../CONTRIBUTING.md)

sdegroot:
Closes #598.

## Problem
On a stencil's detail page, the "Templates using this stencil" table only let you bulk-select template **drafts** for a stencil-version upgrade. A **published** template with no open draft was greyed out and un-upgradeable, so there was no way to roll a published template onto a newer stencil version.

## What changed

**Upgrade published templates.** `UpdateStencilInTemplate` ensures the variant has a draft first — creating one seeded from the latest published version — then upgrades the stencil in that draft. The live published version is untouched until you publish the new draft. The confirmation dialog and a note above the table explain this.

**Per-variant selection.** Exactly one row per variant is actionable (its draft, else its latest published version). Non-upgradable rows show an **info icon** whose tooltip says why (subscribed/read-only catalog, the variant already has a draft, or a newer version supersedes it).

**Filter + pagination (plain HTMX).** A 3-way **filter** select (All / Upgradable / Not upgradable) and **pagination** at 100 rows/page — both `hx-get` swapping `#usage-list`, no JS. The bulk upgrade itself stays client-side JS and refreshes the current filter/page when it finishes. Capped at **100 templates per run**.

**Versions table "Uses" column.** Each stencil version shows how many times it's embedded across draft/published templates.

**Archive left non-destructive.** Archiving only removes a version from the insertion picker (embedded content is inlined into templates), so no hard block; the confirm text was clarified. Editing a stencil whose versions are **all archived** no longer errors — `CreateStencilVersion` falls back to the latest version of any status.

**Removed** the redundant "New Draft" button on the stencil page (versions are created by editing in the editor).

UI-only capability: the `/upgrade` endpoint is UI-only (the editor's in-component upgrade is client-side; no REST/MCP equivalent), so no cross-surface drift.

## From review (this PR)
- **Atomic transaction.** `UpdateStencilInTemplate` previously created the draft via a separate mediator dispatch (its own transaction) then upgraded in a second one. Extracted draft-creation into a `Handle`-scoped `DraftVersionFactory` shared by `CreateVersion` and `UpdateStencilInTemplate`, so create-draft + upgrade now commit/roll back together.
- **DB-side filter/pagination.** The per-variant `upgradable` flag, block reason, filter and pagination are now computed in SQL (`GetStencilUsagePage`, window functions + `LIMIT/OFFSET` + a counts query) instead of loading the whole usage set and slicing in memory; the "Uses" column uses a dedicated `CountStencilUsageByVersion` aggregate.
- CodeQL "unread variable" finding resolved by the atomic-transaction refactor.

## Testing
- `StencilIntegrationTest` — published-template upgrade creates a draft pinned to the new version while the published version stays on the old one; per-variant `upgradable`/block-reason; SQL page counts; all-archived reopen.
- `StencilHandlerHtmxTest` — published-template upgrade returns 200 (was 400); `filter=upgradable` / `filter=not-upgradable`; Uses column renders.
- All CI green (CodeQL, Unit & Integration, Playwright UI). ktlint + `pnpm format:check` clean.

## Follow-ups (filed, not in this PR)
- #612 — preview dropped overrides (dry-run impact) before a bulk upgrade.
- #613 — surface stencil upgrades in the editor without a discard + page refresh.


mvanleijenhorst (approved):
looks good

Issue #615, opened 2026-06-23
Editor: support headless (non-visible) plugin contributions — shortcuts + overlay surface


sdegroot:
## Context

The editor plugin system (`modules/editor/src/main/typescript/plugins/types.ts`) currently only supports **visible** contributions: `sidebarTab` (an always-visible tab) and `toolbarActions` (an always-visible button). A plugin cannot contribute a feature that is *hidden* — e.g. reachable only via a keyboard shortcut, with no toolbar chrome.

This came up while adding a power-user "inspect effective template JSON" debug viewer (`Leader + J`). Because the requirement was explicitly *shortcut-only, no toolbar button*, it could not be modelled as a plugin and was instead shipped as a self-contained core component (`ui/EpistolaTemplateJsonViewer.ts`) with its command registered in the core shortcut registry. That is the right call for one internal tool, but the underlying gap remains: **there is no way to build a headless plugin.**

## What is missing

Two pieces of framework wiring, both partially scaffolded already:

1. **Plugin shortcut contributions are validation-only — not wired into the runtime.**
   `definePluginShortcutContribution` / `mergeShortcutRegistries` exist in `shortcuts/plugin-registry.ts` and are exported from `lib.ts`, but the editor builds its runtime resolver from only the two **core** registries:
   ```
   // EpistolaEditor.ts
   mergeRegistries(getEditorShortcutRegistry(), getInsertDialogShortcutRegistry())
   ```
   Plugin shortcut contributions are never merged in, and they are never fed into the shortcuts-help projection (`ui/shortcuts.ts`). So a plugin-declared shortcut neither fires nor appears in the `Leader + ?` help dialog.

2. **No plugin-owned overlay/surface.** Plugins can only render into `sidebarTab` / `toolbarActions`. A headless plugin that wants to show a popover/dialog/overlay has nowhere to draw it. We need a new SPI — e.g. an `overlays?` contribution the host renders into a dedicated layer — handed the same `PluginContext`.

## Proposed work

- [ ] Merge plugin shortcut contributions into the runtime resolver **and** the shortcuts-help projection (so plugin shortcuts fire and are discoverable). This is the cheaper, already-scaffolded half.
- [ ] Add a host-rendered overlay/surface SPI so a plugin can render a free-floating popover/dialog without a toolbar/sidebar slot.
- [ ] Allow a fully headless plugin: only `init()` + a shortcut contribution + an overlay, no visible `sidebarTab`/`toolbarActions`.
[...]

PR #617, opened 2026-06-23
feat(editor): JSON inspector (data example + template model) with Leader shortcuts

sdegroot:
## Summary

Adds a power-user **JSON inspector** to the template editor — view the current data example *or* the effective template model (`engine.doc`) — and fixes two related editor shortcut/toolbar issues found along the way.

### What you get
- A toolbar **JSON inspector**: a read-only, copyable popover (pinnable + draggable) with a segmented **Template | Data** toggle.
  - **Template** view serializes `engine.doc` (the authored node/slot graph that is saved and sent to preview), kept live as you edit.
  - **Data** view shows the current data example (tab enabled only when the template has examples).
- Shortcuts: **`Cmd/Ctrl + .` then `E`** opens the Data view, **`… then J`** opens the Template view. Both appear in the shortcuts help (`Leader + ?`).

## Commits

1. **feat: add Leader+J effective-template-JSON viewer** — initial command + viewer, registered in the core shortcut registry.
2. **fix: always render keyboard-shortcuts trigger** — the `?` trigger (and the whole right toolbar cluster) was nested inside the data-example selector, so it vanished on templates with **no data examples**, leaving no way to open the shortcuts help. It now renders in an always-present, right-aligned cluster.
3. **refactor: unify data/template into a dedicated `<epistola-json-inspector>`** — extracts the inspector (and ~300 lines of pin/drag/copy/state) out of `EpistolaToolbar` into a self-contained component with its own `json-inspector.css`; the toolbar just hosts it and delegates the open shortcuts.
4. **fix: move leader key to `Cmd/Ctrl + .`** — was `Cmd + Space` (macOS Spotlight). `Ctrl + Space` is also OS-bound on macOS (input-source switch), and the resolver can't distinguish Cmd from Ctrl (both `mod`), so the leader moved off Space to `.`, free of default OS bindings on macOS/Windows/Linux.

## Notes
- Pure editor-module change — no backend/REST/MCP surface or catalog impact (a client-side debug view of in-memory state).
- A follow-up to let this kind of feature live as a *headless plugin* (the plugin system currently only supports visible contributions) is tracked in #615.

## Testing
- `pnpm --filter @epistola/editor test` — **1784 passing** (new `EpistolaJsonInspector.test.ts`, updated toolbar/registry/projection tests).
- `tsc -b` clean; `pnpm format:check` clean.
- Editor UI tests (`EditorShortcutsUiTest`) updated to drive the leader via `Control+Period` and assert the inspector tabs/views.


mvanleijenhorst:
Code looks good, only if I test is it seems not to work. But maybe that is on my side.

mvanleijenhorst (approved):
Code looks good, or I look at the wrong place or it doesn't work on my machine.
If it works on your machine, it fine and PR is approved

sdegroot:
> Code looks good, or I look at the wrong place or it doesn't work on my machine.
> 
> If it works on your machine, it fine and PR is approved

Did you run pnpm build?

whit3st (approved):
It works as intended. A question: should this be always available or gated, for example development only? Would this be useful for the end user is what i am wondering.

sdegroot:
its mostly for us - when there is troubleshooting - we currently have an issue for which i'd like to have this

PR #626, opened 2026-06-24
fix(generation): page header/footer auto-grow to fit content

### LINKED ISSUE #420: Flow-anchored running headers (single-pass, dynamic height)
## Summary

Replace the shipped **positional ≤2-header** model (#399 — ≤2 `pageheader` nodes, both direct children of root; first → page 1, second → pages 2..N) with a **flow-anchored running header**:

- A `pageheader` node becomes a **flow anchor** placeable anywhere in content flow, unlimited count.
- It still paints in a reserved band at the **top** of every governed page (drawn at iText `END_PAGE`); the anchor only selects *which* header fills the band, from its position onward.
- Header height is **dynamic** — measured from the header's own content (no `height` prop). Each governed page reserves a **tight** band equal to its active header's measured height.

This is a **separate follow-up change** (branch fresh from `main`); the shipped branch/PR #399 is untouched.

## Why it resolves in a single pass

Page N's band depends only on which header is active *entering* page N, which depends only on anchors that landed on pages < N — all already committed by the time iText lays out page N. So it resolves in **one forward layout pass** via a custom `DocumentRenderer` per-page hook. No fixed point, no convergence machinery, no extra header pass. (A second pass only occurs if the template already uses the pre-existing, unrelated `sys.pages.total` two-pass, which the header mechanism free-rides on.)

## Confirmed design decisions

1. **Scope**: new follow-up change from `main`; shipped branch untouched.
2. **Height**: dynamic — measured from content; `height` prop/inspector field removed. Tight per-page band.
3. **Model**: marker/running-header — header drawn in the top band; anchor selects which.
4. **Anchor in loop/datatable**: allowed silently, first-class (one promotion per rendered occurrence).
5. **Switch timing**: takes effect **on the page after** the anchor's landing page ("the running header is the section you were in *entering* the page"). Anchors before any content (declared at document top) are active from page 1 → reduces exactly to today's single-header behavior.
6. **Passes**: the header feature adds **zero** extra passes.

## Scope (all layers — single source of truth going forward)

- **Renderer**: new `AnchorMarkerElement`, `AnchorLandingSink`, `BandedDocumentRenderer`, pure `PageHeaderScheduleRule`; rework `DirectPdfRenderer`/`PageHeaderEventHandler`/`PageHeaderNodeRenderer`/`RenderContext`; delete the `firstPageSpacer`/`computeHeaderBands` page-1 hack.
- **Editor**: registry (drop `maxInstancesPerDocument:2`, remove `height` field, narrow `isAnchoredPageBlock` to footer, add `isFlowAnchor`, new examples); commands/DnD (header placeable anywhere); minimal inline-anchor UX.
- **Backend**: delete `PageHeaderCardinalityValidator` + its `UpdateDraft` wiring/tests.
- **MCP**: `pageheader` `maxInstancesPerDocument` → `null`.
- **Demo catalog**: add a mid-flow running-header demo template; bump catalog version.
- **Docs**: rewrite `docs/generation.md` header section; new ADR `docs/adr/0003-flow-anchored-running-headers.md`; CHANGELOG.

## Full design

See [`docs/plans/flow-anchored-running-headers.md`](https://github.com/epistola-app/epistola-suite/blob/main/docs/plans/flow-anchored-running-headers.md) — switch rule (with worked example), single-pass architecture, testing strategy, risks.

Predecessor: #399.

sdegroot:
Closes #420 — supersedes the flow-anchored running-header design with a smaller auto-grow approach (height stays a *minimum* rather than being removed). See the issue comment for the scope intentionally not done.

## Problem

Changing a template's **page-header height** sometimes had **zero effect** — and on a real customer template the whole letterhead vanished.

Root cause: a page header/footer renders its content into an iText `Canvas` clipped to a **fixed rectangle of exactly its configured `height`**. Content taller than that is **dropped** by iText, leaving a blank band. It only triggers when header content exceeds the set height (a letterhead with a logo, an address block, several lines), so it was hard to reproduce with a plain-text header and looked like "I change the height and nothing happens."

Ruled out first: not a cache, not a save problem (`props.height` round-trips faithfully), not the band math.

## What's in this PR

Three related fixes, then a cleanup + docs:

### 1. Auto-grow — `height` is now a minimum
- Pre-measure each header/footer's natural content height via a discarded **iText dry-layout** pass (its own `FontCache`, since iText fonts are document-bound).
- Reserve `max(configured height, content height)` for **both** the drawn band and the body top/bottom margin, so they always agree.
- `OVERFLOW_Y/X = VISIBLE` as a safety net; measurement runs only when a header/footer exists and falls back to the configured height on any error, so it can never break a previously-working render.
- Net: increasing `height` adds whitespace; content shorter than it is unaffected; content **taller** than it grows the band instead of vanishing.

### 2. A nested address block no longer inflates the band
An `addressblock` is a page-absolute element that emits a ~window-height spacer meant for the **body**. The header rendered the un-hoisted graph, so that spacer landed inside the header and (with auto-grow) ballooned it. Fix: hoist the address block to the body root **before** the bands are measured/rendered, so the header never contains it.

### 3. The address reservation respects the header height
The address block's body-side reservation (space so text clears the recipient window) was derived from the **raw** `height`, so a tall auto-grown header still over-reserved. The renderer now passes the resolved page-1 body top via `RenderContext.bodyContentTopPt`; the reservation shrinks — to zero under a header that already covers the window. (Measured on the customer template: body moved up from 363pt to 267pt = the window bottom.)

### 4. Refactor (no behaviour change)
Hoist the address block **once** per render path and pass the hoisted graph down (was hoisting twice under two names); bundle measure-heights + bands + footer-height into `resolveBandLayout()` so the single- and two-pass paths stop duplicating orchestration.

### 5. Docs
`docs/generation.md` gains **"Header & footer band height (auto-grow)"** and **"Address blocks vs. header/footer bands"** subsections (the measure→resolve→render flow, the hoist-before-band rule, the `bodyContentTopPt` reservation), each with a code map.

## Key files

- `HeaderFooterBand.kt` (new) — `buildBandWrapper()` (shared by the handlers **and** measurement, so measured == rendered) + `measureBandContentHeight()`.
- `DirectPdfRenderer.kt` — `measureEffectiveBandHeights`, `resolveBandLayout`, `computeHeaderBands`, `hoistAddressBlock` (hoist-before-band), `bodyContext`.
- `PageHeaderEventHandler` / `PageFooterEventHandler` — effective-height map + shared builder + overflow-visible.
- `AddressBlockNodeRenderer` + `RenderContext.bodyContentTopPt` — header-aware reservation.

Each of the (up to two) headers and the footer is measured **independently, once per render** (not per page); the event handlers look up the precomputed height on each `END_PAGE`.

## Testing

- `PageHeaderFooterTest`: tall content still renders (the bug); band auto-grows so the body clears it; footer parity; configured height honoured as a minimum; address-in-header doesn't inflate the band; reservation shrinks to zero under a tall header.
- Verified end-to-end against the real customer template (rendered through the preview path, text positions measured).
- Full `:modules:generation:test` green; `ktlintCheck` + `pnpm format:check` clean.

## Not included (follow-ups)

- Editor: relabel the height field to "Min height" + the `sp` base-unit fix.
- Two-header band coupling for the cover-shorter-than-running case (still additive page-1 spacer).
- Demo-catalog coverage for a tall-content header.
- The address block ideally shouldn't be authored inside a header at all (its fixed window can still geometrically overlap a tall one) — see linked discussion.


modules/generation/src/main/kotlin/app/epistola/generation/pdf/DirectPdfRenderer.kt
```diff
@@ -415,13 +420,18 @@ class DirectPdfRenderer(
         iTextDocument.setFont(context.fontCache.regular)
         iTextDocument.setMargins(topMargin, rightMargin, bottomMargin, leftMargin)
 
+        // `document` arrives address-block-hoisted from the caller: the address block
```
mvanleijenhorst: **Implicit "already hoisted" precondition on `performRenderWithContext`.** The parameter is named `document` but must already have gone through `hoistAddressBlock` — enforced only by this comment. Both current callers (single- and two-pass) do it correctly, but a future third caller passing a raw graph would silently reintroduce the "address block inflates the band" bug, with no compile or test failure. Consider making the invariant structural: rename the param to `hoistedDocument`, or add a cheap assert that no `addressblock` is a descendant of a band node.

modules/generation/src/main/kotlin/app/epistola/generation/pdf/DirectPdfRenderer.kt
```diff
@@ -684,6 +701,128 @@ class DirectPdfRenderer(
...
+        context: RenderContext,
+        pageSettings: app.epistola.template.model.PageSettings,
+        renderingDefaults: RenderingDefaults,
+        pdfaCompliant: Boolean,
+        fontFamilyResolver: FontFamilyResolver?,
+    ): Map<String, Float> {
+        if (headerNodes.isEmpty() && footerNode == null) return emptyMap()
+
+        val measureContext = context.copy(fontCache = FontCache(pdfaCompliant, fontFamilyResolver))
+        val pdfDocument = PdfDocument(PdfWriter(OutputStream.nullOutputStream()))
```
mvanleijenhorst, line 777: **Possible resource leak if measurement setup throws.** `pdfDocument`, `iTextDocument`, and `createDefaultRegistry(pdfDocument)` are created *before* the `try/finally`. If `getPageSize`, the `Document` constructor, `setFont`, or `createDefaultRegistry` throws, `pdfDocument` is never closed (its `PdfWriter`/streams leak). Low probability for a valid `pageSettings`, but worth tightening — pull the creation inside the `try`, or wrap with `.use {}`, so cleanup always runs.

modules/generation/src/test/kotlin/app/epistola/generation/pdf/PageHeaderFooterTest.kt
```diff
@@ -623,6 +623,210 @@ class PageHeaderFooterTest {
         return extractFirstBaselineY(pdfBytes, "FOOTER CONTENT")
     }
 
+    // -----------------------------------------------------------------------
+    // Auto-grow: a header/footer band reserves max(configured, content) height,
```
mvanleijenhorst, line 627: **Test gap: the two-pass render path is not exercised with tall band content.** All auto-grow tests here use plain-text docs (no total-pages token), so they run through the single-pass path. This PR refactored *both* paths, and `renderTwoPass` has its own orchestration that threads `effectiveHeights` into two separate passes — a regression there (e.g. `effectiveHeights` not forwarded to pass 2) would not be caught. Adding one test with a tall band **and** a `total pages` token (to trigger two-pass) would close this.

modules/generation/src/test/kotlin/app/epistola/generation/pdf/PageHeaderFooterTest.kt
```diff
@@ -623,6 +623,210 @@ class PageHeaderFooterTest {
...
+        )
+    }
+
+    private fun bodyStartY(doc: TemplateDocument): Float {
+        val pdfBytes = ByteArrayOutputStream().also { renderer.render(doc, emptyMap(), it) }.toByteArray()
+        return extractFirstBaselineYOnPage(pdfBytes, "BODYSTART", 1)
+    }
+
+    @Test
+    fun `header content taller than the configured height still renders (not clipped)`() {
```
mvanleijenhorst, line 675: **The safety nets that back the "can never break a working render" guarantee are untested.** The measurement fallback (`catch → log.warn → configured height` in `measureEffectiveBandHeights`) and the `OVERFLOW_Y/X = VISIBLE` canvas property are the two mechanisms ensuring auto-grow can't make a previously-working render fail — but neither has a test. The fallback is admittedly hard to force (you'd need iText layout to throw), but it's exactly the guarantee the PR leans on, so it's worth at least one targeted case if it can be provoked.

docs/generation.md
```diff
@@ -243,6 +243,75 @@ Out of scope for this iteration (filed as follow-ups): last-page header,
 per-section / page-range headers, odd/even alternating headers, and the same
 variant model for footers.
 
+### Header & footer band height (auto-grow)
```
mvanleijenhorst, line 246: **Consider an ADR for the user-facing semantic change.** This is great module-level documentation of the *mechanism* (measure → resolve → render, hoist-before-band, `bodyContentTopPt`) — the right home for that. What it doesn't capture is the *decision*: `height` now means "minimum" rather than "fixed/clip", which is a user-facing semantic change with real alternatives (clip + warn, reject overflow, manual height) and downstream consequences (the editor "Min height" relabel, two-header coupling, address-block-in-header). The repo has a `docs/adr/` convention (0001–0007) for exactly this kind of decision, and the PR description already contains ADR-shaped reasoning ("Ruled out first: not a cache, not a save problem…") that will otherwise be lost once the PR scrolls out of view. A short ADR (Context / Considered options / Decision / Consequences) would anchor it. If the team's bar keeps ADRs for cross-module/wire-format decisions only, an explicit "module-doc, no ADR because …" note is also fine — just make the call deliberately.

mvanleijenhorst (commented):
## Review: architecture, tests, ADR

Strong, well-reasoned fix. The core design — `height` becomes a *minimum* via a dry-layout pre-measurement, with a single shared `buildBandWrapper` so that *measured == rendered* — is the right architecture, and the de-duplication of the two event handlers is a real improvement. The shared `FontCache`-per-measurement-document handling correctly avoids font leakage. No blocking issues. Inline comments below cover a few robustness/test points plus the ADR question.

Note: the body-reservation reference point (`bodyContentTopPt`) is always set on the production render path (`DirectPdfRenderer` before body layout), so the raw-height fallback in `AddressBlockNodeRenderer` is effectively test-only and is intentionally fixed/raw-based — not flagging it.

modules/generation/src/main/kotlin/app/epistola/generation/pdf/DirectPdfRenderer.kt
sdegroot: Done in 5dbd2b7c — renamed the param to `hoistedDocument` with a KDoc note stating the precondition. Kept it a rename rather than re-hoisting inside (that would undo the "hoist once per path" cleanup); both callers hoist, and a future caller now has an explicit signal.

modules/generation/src/main/kotlin/app/epistola/generation/pdf/DirectPdfRenderer.kt
sdegroot: Good catch — fixed in 5dbd2b7c. The throwaway `PdfDocument` is created just before a `try` that now wraps all of `Document`/`setFont`/`createDefaultRegistry` + the measurement, and the `finally` closes the `PdfDocument`, so a setup failure can't leak the writer/streams.

modules/generation/src/test/kotlin/app/epistola/generation/pdf/PageHeaderFooterTest.kt
sdegroot: Added in 5dbd2b7c: `auto-grow works on the two-pass render path` — a `sys.pages.total` expression forces two-pass (asserted via `TwoPassAnalyzer.requiresTwoPassRendering`), then checks the tall header content still renders. Guards the `effectiveHeights` threading through both passes of `renderTwoPass`.

modules/generation/src/test/kotlin/app/epistola/generation/pdf/PageHeaderFooterTest.kt
sdegroot: Agreed these are the guarantee, and I looked at testing both — they're defensive-only and not deterministically provokable through the public API:
- **Measurement fallback**: only fires on an *unexpected* iText layout throw. The realistic trigger (a missing asset) throws only in STRICT mode — but then the real render throws on the same node too, so the fallback can't actually rescue it; in PREVIEW mode the asset degrades to a placeholder and nothing throws. Forcing it would need fault injection into the private `createDefaultRegistry`/measurement path.
- **OVERFLOW_VISIBLE**: can't be reached in normal flow because the reserved height is already `max(configured, content)`, so the canvas never clips. It's pure belt-and-suspenders for a measurement/round-off edge.

So I left both untested deliberately rather than add a contrived fault-injection harness. Open to revisiting if we add a seam that makes the failure path injectable.

docs/generation.md
sdegroot: Done in 5dbd2b7c — added [ADR 0008](docs/adr/0008-header-footer-height-minimum.md) (Context / Considered options [clip, warn, reject-at-save, auto-grow, always-auto-size] / Decision / Consequences), and cross-linked it from the `docs/generation.md` section. The module doc keeps the mechanism; the ADR captures the decision.

PR #627, opened 2026-06-24
feat(templates): sortable, paginated templates list

whit3st:
## Description

Adds server-rendered **sorting** and **pagination** to the Document Templates list page (`templates/list.html`) — scoped to that page only, no shared data-table toolkit.

- **Sorting:** clickable column headers (Name, Catalog, Variants, Last Modified). Click to sort, click again to flip direction. Backed by a whitelisted `TemplateSort` enum so the `ORDER BY` column is never raw input; default remains most-recently-modified first.
- **Pagination:** 10 rows per page with Previous / "Page _n_ of _m_" / Next; the bar appears only when there's more than one page. A new `CountTemplateSummaries` query supplies the total.
- **Mechanics:** both `list` (full page) and `search` (HTMX fragment) share one model builder. Sort/page controls swap the whole table wrapper via HTMX (so the active-sort indicator and pager update together) and `hx-include` the live search box; a plain `th:href` is a no-JS fallback.
- **Sort icons:** every sortable column **always** renders exactly one chevron so the header never shifts width — a faint neutral hint when inactive, a solid directional chevron (`chevron-up` asc / `chevron-down` desc) when active. Added `chevron-up` to the design-system icon sprite/allow-list.

## Type of Change

## Component(s) Affected

## Checklist

## Additional Notes

- Page size is **10**; with the 4-template demo catalog you won't see the pagination bar until a tenant has >10 templates (the sort headers always show).
- `ListTemplateSummaries` gained `sort`/`descending` params with defaults, so the existing **MCP** caller is unaffected.
- New/updated tests in `DocumentTemplateRoutesTest` cover paging (page 1 / last page), name sort asc & desc, and assert all 4 sortable columns always carry exactly one chevron.


modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/queries/ListTemplateSummaries.kt
```diff
@@ -94,3 +121,45 @@ class ListTemplateSummariesHandler(
...
+            }
+        }
+
+        val jdbiQuery = handle.createQuery(sql)
+            .bind("tenantId", query.tenantId.key)
+        if (query.catalogKey != null) {
+            jdbiQuery.bind("catalogKey", query.catalogKey)
+        }
+        if (!query.searchTerm.isNullOrBlank()) {
+            val escaped = query.searchTerm.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
```
mvanleijenhorst: **Duplicated ILIKE-escape logic (drift risk).** This `replace(...)` chain, the `ESCAPE '\'` clause, and the null/blank guards are copy-pasted verbatim from `ListTemplateSummariesHandler` (line 114). The two queries **must** stay byte-identical — if they ever drift, the pager's "Page _n_ of _m_" will count a different set than the table renders (e.g. count says 2 pages, but page 2 comes back empty). Suggest extracting a shared helper (e.g. `private fun escapeLike(term: String)` plus a small `Handle.bindNameFilter(...)`) so the invariant is structural rather than a matter of discipline. Highest-value cleanup in the PR.

apps/epistola/src/main/resources/templates/templates/list.html
```diff
@@ -30,53 +30,119 @@
...
+                            </td>
+                        </tr>
+                    </tbody>
+                </table>
+
+                <nav th:if="${totalPages > 1}" class="ep-pagination" aria-label="Templates pagination" data-testid="template-pagination">
+                    <a th:if="${hasPrev}"
+                       class="ep-btn ep-btn-ghost ep-btn-sm"
+                       data-testid="pagination-prev"
+                       th:href="@{/tenants/{tenantId}/templates(tenantId=${tenantId},page=${prevPage},sort=${sort},dir=${sortDir},catalog=${selectedCatalog})}"
```
mvanleijenhorst, line 118: **No-JS fallback silently drops the active search.** This pagination `th:href` (and the Next link below, and the `sort-header` `th:href`) intentionally omit `q`. With JS, `hx-include="[name='q']"` preserves the filter; without JS, clicking Prev/Next or a column header reloads the **full unfiltered list**, discarding what the user typed. Compounding it, the search `<input>` in `fragments/search.html` has no `th:value="${searchTerm}"`, so a server-rendered filtered page shows results with an empty box. If no-JS isn't a supported mode you can ignore this — but then these `th:href` fallbacks are decorative and could be dropped.

apps/epistola/src/main/resources/templates/templates/list.html
```diff
@@ -30,53 +30,119 @@
...
+                </nav>
+            </th:block>
+        </div>
     </th:block>
+
+    <!--/* Sortable column header. Clicking re-sorts via HTMX (swapping the whole
+           table wrapper so the active-sort indicator and pagination update too);
+           the th:href is a no-JS fallback that reloads the full list page. */-->
+    <th th:fragment="sort-header(label, col)"
+        th:with="active=${sort == col},nextDir=${(sort == col and sortDir == 'asc') ? 'desc' : 'asc'}"
```
mvanleijenhorst: **`defaultDescending` is effectively dead for inactive columns.** `nextDir` always resolves to `'asc'` for any column that isn't the active one, so the *first* click on Variants or Last Modified sorts **ascending** (fewest variants / oldest first) — not the "natural" descending that `TemplateSort.defaultDescending` documents. That property only ever fires for the initial `sort=updated` default load. Either honor it here (inactive column → its `defaultDescending`) or drop the property so it doesn't read as a live rule.

apps/epistola/src/main/resources/templates/templates/list.html
```diff
@@ -30,53 +30,119 @@
...
-            <div class="empty-state-description">Create your first document template to get started.</div>
-            <a th:if="${auth.has('TEMPLATE_EDIT')}"
-               th:href="@{/tenants/{tenantId}/templates/new(tenantId=${tenantId})}" class="ep-btn ep-btn-primary" style="margin-top: var(--ep-space-4);">
-                <th:block th:replace="~{epistola-web/icon :: icon(name='plus', class='ep-icon')}" />
-                New Template
-            </a>
-        </div>
+        <div id="template-table">
+            <th:block th:fragment="table">
+                <div th:if="${total == 0}" class="empty-state">
```
mvanleijenhorst, line 43: **Empty-state copy is wrong under a catalog filter.** `total == 0 && searchTerm == ''` always renders "No templates yet / Create your first…", but this branch also fires when a **catalog filter** (chosen from the dropdown, not a search) legitimately matches zero templates. A tenant with plenty of templates in other catalogs then sees a misleading "get started" CTA. Consider treating an active `selectedCatalog` like an active search here (a "no templates in this catalog" message).

mvanleijenhorst (commented):
First of all it looks great!

## Review: architecture, code quality, UX

Solid, ship-able change — clean CQRS fit (`TemplateSort` whitelist + sibling `CountTemplateSummaries`), a shared model builder feeding both the full page and the HTMX fragment, and good a11y details (always-one-chevron, `aria-sort`, `aria-disabled`, `aria-hidden` icons). No correctness bugs that crash or corrupt. Inline comments below cover one drift risk and three UX/dead-code points. Two notes that aren't line-specific:

**Test coverage gap** (`DocumentTemplateRoutesTest`): the new tests cover paging (first/last) and name asc/desc, but not sort by `variants`/`catalog`/`updated`, the invalid-`sort`-param fallback to `UPDATED`, or catalog-filter + pagination together. Since the sort column is now user-driven, a couple more cases would be worthwhile.

**Architecture / scope**: sorting + pagination is deliberately bespoke to this one page (the PR says so). Reasonable as a first cut. The shared CSS (`.ep-sort`, `.ep-pagination`) landed in the design system, but the handler/template wiring is copy-forward-able — worth flagging as known debt so that when a second list page (themes/stencils/images/tenants/environments) needs the same thing, it gets extracted into a shared toolkit rather than re-pasted.

apps/epistola/src/main/resources/templates/templates/list.html
whit3st: Ah, this is an existing issue it seems, good catch! I'll fix it here as might well and we should also check other forms that have filters I think.

PR #663, opened 2026-07-01
fix(test): warm up logging once before parallel tests run

sdegroot:
## Problem

Under class-level test parallelism (many test classes starting in the same JVM at once), the **first** classes to touch a logger can catch SLF4J still finishing its one-time startup. During that split-second SLF4J hands back a **placeholder** logger instead of the real Logback one, which caused two intermittent, parallel-only symptoms:

- `ClassCastException` — the placeholder can't be cast to a Logback `Logger`.
- Log events **silently dropped** — they go to the placeholder and never reach the real appenders.

## Fix

Finish SLF4J's startup **once, single-threaded, before any test runs**. The JVM-wide `TestRuntimeLifecycleListener` (already present) now completes the SLF4J→Logback binding at launcher-session open — before any test class is instantiated — so no parallel test can observe the half-initialized state:

```kotlin
override fun launcherSessionOpened(session: LauncherSession) {
    LoggerFactory.getILoggerFactory()
}
```

Parallel execution is unchanged. Two files: the listener + a CHANGELOG entry.

## Verification

- `ktlintCheck` clean.
- **4× consecutive full `unitTest`** runs green. Before the fix, a full parallel run failed ~5 tests (`HeartbeatFailureLogTest`, `SpringAiMcpNoiseRewriterTest`) with exactly the `SubstituteLogger → Logger` `ClassCastException` and dropped-event symptoms above.


modules/epistola-core/src/main/kotlin/app/epistola/suite/logs/ApplicationLogIngestor.kt
```diff
@@ -145,30 +143,29 @@ class ApplicationLogIngestor(
     }
 
     private fun attachAppender() {
-        val loggerContext = LoggerFactory.getILoggerFactory() as? LoggerContext ?: run {
+        // Attach/detach on the shared root logger goes through the hardened, synchronized,
+        // substitute-init-safe seam (LogbackAppenders) — never the raw `getILoggerFactory()`
+        // cast, which races under the parallel test suite.
+        val attached = LogbackAppenders.attachRootAppender(APPENDER_NAME) { loggerContext ->
+            val threshold = ThresholdFilter().apply {
+                setLevel(Level.toLevel(properties.level, Level.INFO).levelStr)
```
sdegroot: does this mean we can only ever log on info level?

modules/epistola-core/src/main/kotlin/app/epistola/suite/logs/LogbackAppenders.kt
```diff
@@ -0,0 +1,117 @@
...
+ *    lock Logback uses for reconfiguration — so a parallel attach/detach can no longer
+ *    corrupt the appender list.
+ *
+ * [attachRootAppender] is **idempotent by name**: an appender already registered under the
+ * same name is detached and stopped first, so two overlapping contexts (e.g. two
+ * `epistola.logs.enabled=true` test classes) end up with a single, last-wins appender on
+ * root rather than a growing stack. Each owner still detaches its own reference on
+ * shutdown via [detachRootAppender].
+ */
+object LogbackAppenders {
```
sdegroot: all of these things are i the main sources. But we only need this stuff for the tests right?

modules/epistola-support-telemetry/src/main/kotlin/app/epistola/suite/support/telemetry/TelemetryLeg.kt
```diff
@@ -150,33 +149,33 @@ class TelemetryLeg(
...
-            return
-        }
-        val threshold =
-            ThresholdFilter().apply {
-                setLevel(Level.toLevel(props.logLevel, Level.INFO).levelStr)
-                start()
+        // Attach/detach on the shared root logger goes through the hardened, synchronized,
+        // substitute-init-safe seam (LogbackAppenders) — never the raw `getILoggerFactory()`
+        // cast, which races under the parallel test suite.
+        val attached =
```
sdegroot: why do we need this actually?

modules/epistola-core/src/main/kotlin/app/epistola/suite/logs/ApplicationLogIngestor.kt
sdegroot: No — and this code is removed anyway (the PR is now simplified to a one-line logging warm-up). For the record: that line set the DB-capture appender's *threshold* from `epistola.logs.level` (default `INFO`, but configurable to ERROR/WARN/INFO/DEBUG/TRACE) via `Level.toLevel(properties.level, Level.INFO)` — INFO is only the fallback if the property is unset/unparseable. A ThresholdFilter captures everything *at or above* that level, and only for the `application_log` table (console/other appenders untouched). It was pre-existing behaviour, just moved into a lambda.

modules/epistola-core/src/main/kotlin/app/epistola/suite/logs/LogbackAppenders.kt
sdegroot: Good instinct — you were right it was over-built. This whole `LogbackAppenders` seam is now removed. It turned out we didn't need it: every flake we actually reproduced was SLF4J caught mid-initialization, fixed by a single one-time logging warm-up before tests run. The appender-mutation race this guarded against never actually showed up, so it's gone.

modules/epistola-support-telemetry/src/main/kotlin/app/epistola/suite/support/telemetry/TelemetryLeg.kt
sdegroot: We don't — removed too. It only existed to route TelemetryLeg through the (now-deleted) seam and satisfy a guardrail that no longer exists. TelemetryLeg is back to its original code, untouched by this PR. The whole fix is now just the one-time logging warm-up in TestRuntimeLifecycleListener.

PR #700, opened 2026-07-05
refactor(chart)!: chart-hygiene pass (secrets-only, remove cnpgExisting.username, schema, grafana fixes)

sdegroot:
A hygiene pass on `charts/epistola` driven by a multi-dimension audit (DB/migration helpers, values, templates, observability, security, docs). Six focused commits; three carry breaking changes (pre-GA, deliberate).

## Commits
1. `fix(chart)!` **Remove `database.cnpgExisting.username`** — it silently connected the app as the CNPG *owner* (CNPG's `jdbc-uri` embeds owner creds, and pgjdbc lets URL creds override the explicit username — verified). Two-role on CNPG is done with `database.type=external`; `migration.credentials` + `cnpgExisting` now fails the render.
2. `refactor(chart)` **Batch-A helper hygiene** — delete the dead `cnpgExisting` branch in `migrationDatabaseEnv`; shared `epistola.database.externalUrl` helper with a **required-host guard** (external + empty host now fails render instead of a broken `jdbc://:5432/` URL); fold supported-type validation into `effectiveType`; `serviceAccount.automount: false` (the app makes no k8s API calls); stale-comment fixes.
3. `feat(chart)!` **Secrets only via `existingSecret`** — remove inline `oidc.clientSecret`, `keycloakAdmin.clientSecret`, `encryption.keys[].material`; the chart never places secret material in the pod spec, values, or the release. Encryption stays enabled-by-default (opt out with `enabled: false`).
4. `fix(chart)` **Grafana** — `GrafanaAlertRuleGroup.folderRef` was a folder *title*, not a `GrafanaFolder` CR name (alerts never reconciled into the folder); added a `GrafanaFolder` template and referenced it by name. Dropped a dead `datasource` picker variable from all 5 dashboards. Documented the grafana-operator CRD prerequisite.
5. `feat(chart)` **`values.schema.json`** — enums for `database.type` / `migration.mode`, required `external.host`/`existingSecret` for `external`; catches typos (incl. the removed `cnpg` type) before render. Also grouped `datasource.hikari` next to `database`.
6. `refactor(chart)!` **Renames for consistency** — `database.cnpgExisting.secretName` → `existingSecret`, `migration.credentials.passwordKey` → `secretKey`.

## Breaking changes
- `database.cnpgExisting.username` (+ its `existingSecret`/`passwordKey`) removed — use `database.type=external` for two-role CNPG.
- Inline `oidc.clientSecret` / `keycloakAdmin.clientSecret` / `encryption.keys[].material` removed — use `existingSecret`.
- `database.cnpgExisting.secretName` → `existingSecret`; `migration.credentials.passwordKey` → `secretKey`.

**Companion infra PR** (epistola-infra `chart-align-migration-secretkey`) carries the matching `secretKey` rename. It's order-independent — the value is `password`, which is also the chart default, so neither chart version breaks. The live epistola-demo already complies with every breaking change (secrets via `existingSecret`, encryption on, `external` two-role).

## Validation
`helm lint` clean; every path render-tested (external two-role, cnpgExisting single-role, `none`); schema rejects bad `database.type`/`migration.mode` and missing external host; no plaintext secret env renders; grafana dashboards parse as valid JSON.

## Deferred (separate PR)
The bigger **observability extraction** — moving the ~60 KB of grafana CRDs to a subchart or `configmap-from-files`. Not urgent now that the bugs are fixed and the prereq is documented.

### DIFF
--- a/charts/epistola/CHANGELOG.md
+++ b/charts/epistola/CHANGELOG.md
@@ -12,9 +12,15 @@
 - **`serviceAccount.automount` now defaults to `false`.** The app makes no Kubernetes API calls, so its pods no longer auto-mount a ServiceAccount token. Set `true` if you add something that needs API access.
 - **`database.type=external` now fails the render when `database.external.host` is empty** (mirroring the existing password guard), instead of emitting a broken `jdbc:postgresql://:5432/…` URL that only failed at pod startup.
 
+### Fixed
+
+- **Grafana alerts now reconcile into the intended folder.** `GrafanaAlertRuleGroup.folderRef` was fed the folder *title* (`observability.grafana.folder`), but `folderRef` must name a `GrafanaFolder` CR. Added a `GrafanaFolder` template (gated by `observability.grafana.enabled`); the alert group now references it by name, co-located with the dashboards.
+- Removed a dead `datasource` template variable from all five dashboards — panels reference the datasource uid directly, so the picker did nothing.
+
 ### Internal
 
 - DB/migration credential helpers de-duplicated: a shared `epistola.database.externalUrl` helper, `epistola.database.effectiveType` now centralizes supported-type validation, and the dead `cnpgExisting` branch left by the `cnpgExisting.username` removal is gone. No behavior change.
+- Documented the grafana-operator CRD prerequisite for `observability.grafana.enabled`.
 
 ## [0.8.0] - 2026-07-04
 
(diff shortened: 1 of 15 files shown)


sdegroot:
Companion infra PR: epistola-app/epistola-infra#13 (the matching `secretKey` rename).

Issue #742, opened 2026-07-14
Retire the content_store legacy blob layer once #738 backfill is complete everywhere


sdegroot:
Follow-up cleanup to #738. The blob-storage restructure deliberately kept the legacy `content_store` table and a transitional migration layer as a safety net. Once every deployment has completed the backfill (verified), this layer should be removed.

## Preconditions to verify before starting
- `content-backfill.completed` marker present (`app_metadata`) on every installation — i.e. the `ContentBackfillRunner` background task has finished a full pass fleet-wide.
- No `assets` rows with `content_hash IS NULL` anywhere (except genuinely-irretrievable legacy bytes, which should be reconciled or accepted as broken).
- No document requests still resolving via the legacy fallback (i.e. all pre-#738 documents are in `document_content`).
- `epistola.storage.orphaned_blobs{namespace=asset}` steady/near-zero (reaper healthy).

## When is it safe to drop? (upgrade-path safety — READ FIRST)

The data move is a **runtime background task**, not a Flyway migration. Flyway replays every intervening schema migration on any version jump, but it does **not** run the backfill — that only happens when a version *containing* `ContentBackfillRunner` boots and stays up long enough to drain `content_store`. So a naive `DROP TABLE content_store` is unsafe: an operator jumping from a pre-#738 version straight to the drop version would have Flyway drop the table with the backfill never having run → data loss.

**Therefore the drop migration must be self-verifying (refuse, don't lose data), gated on "has un-migrated data" AND "backfill completed":**

```sql
DO $$ BEGIN
  IF EXISTS (SELECT 1 FROM content_store LIMIT 1)
     AND NOT EXISTS (SELECT 1 FROM app_metadata WHERE key = 'content-backfill.completed') THEN
    RAISE EXCEPTION 'content_store still holds un-migrated blobs and the #738 backfill has not completed here. '
      'Run a backfill-bearing version until it finishes (app_metadata ''content-backfill.completed'' set; '
      'epistola.storage.orphaned_blobs stable), then upgrade to this version.';
  END IF;
END $$;
DROP TABLE content_store;
```

Airtight across paths: **fresh install** → table empty → drops cleanly; **upgrade with backfill completed** → rows present (retained safety-net copies) but marker present → drops; **big jump / backfill never ran / mid-flight** → rows present, no marker → migration **fails loudly with remediation**, deploy blocked, no data loss.

## Upgrade-path policy this establishes
[...]

Issue #42, opened 2025-12-26
Implement image block in editor


sdegroot:
## Summary
Add support for inserting images into templates via the visual editor.

## Details
- New ImageBlock type in TemplateModel
- UI for selecting images from asset manager
- Support for sizing, positioning, and alt text
- Link to asset manager for uploads

## Acceptance Criteria
- [ ] ImageBlock added to block types
- [ ] Images can be inserted via editor
- [ ] Image sizing controls available
- [ ] Alt text configurable (for accessibility)
- [ ] Images render in preview and generation

## Technical Notes
- Depends on asset manager for storage
- Consider lazy loading for large images

PR #743, opened 2026-07-14
feat(storage)!: restructure blob storage — reclaim document blobs + content-addressable assets (#738)

### LINKED ISSUE #738: Restructure blob storage: reclaim orphaned document blobs + content-addressable asset store
## Summary

Automatic retention **does not reclaim blob storage**. When the monthly-partition retention drops old `documents` partitions, the corresponding rows in `content_store` (the default Postgres blob backend) are **orphaned and never deleted**, so blob storage grows without bound even though document metadata stays bounded. Under high volume (100k+ documents/day), `content_store` is the dominant storage cost and it leaks every retention cycle.

This issue proposes a **restructure of blob storage** that both fixes the leak and deduplicates the long-lived side of the store.

## Root cause

Two deletion paths, only one of which reclaims blobs:

| Path | `documents` row | `content_store` blob |
| --- | --- | --- |
| Explicit `DeleteDocument` (API/UI delete) | deleted | **deleted** ✅ (`DeleteDocumentHandler` → `contentStore.delete(ContentKey.document(...))`, `documents/commands/DeleteDocument.kt:57`) |
| Automatic retention (after `epistola.partitions.retention-months`, default 3) | partition dropped | **orphaned** ❌ |

Automatic retention never goes through `DeleteDocument`. `PartitionMaintenanceScheduler.dropOldPartitions` (`documents/cleanup/PartitionMaintenanceScheduler.kt`) calls `epistola_drop_partitions_before('documents'::regclass, cutoff)` — a raw `DROP TABLE <child_partition>`. That drop **cannot cascade** to `content_store` because it is **unpartitioned, un-retained** (`db/migration/core/V20260515091100__core_content_store.sql` — `key VARCHAR(512) PK, content BYTEA, …`), a partition `DROP TABLE` fires no per-row delete, and there can be no FK (see below).

The default backend is `POSTGRES` (`storage/StorageConfiguration.kt`, `epistola.storage.backend`); `S3ContentStore` / `FilesystemContentStore` have the **same orphan-on-drop gap**.

## What `content_store` holds (two namespaces)

`storage/ContentKey.kt` defines exactly two key namespaces, with very different lifecycles:

- **`documents/{tenant}/{docId}`** — generated PDFs. Ephemeral, high-churn, **age out** with document retention. Written by `DocumentGenerationExecutor` (`documents/batch/DocumentGenerationExecutor.kt:110`).
- **`assets/{tenant}/{assetId}`** — asset content: logos, images, **font files** (anything in the `asset_types` lookup). Long-lived reference data, low-churn, **keep-forever**. Written by `UploadAsset` (user upload) and `ImportAsset` (catalog import); deleted directly today by `DeleteAsset`.

This is why `content_store` cannot simply be partitioned by `created_at` and drop-retained: a monthly partition would mix droppable document blobs with keep-forever asset blobs (e.g. fonts), and dropping it would destroy the assets.

## Evidence (reproduced empirically)

Confirmed via the `epistola-load-test` `volume` track against a local `1.0.0-RC3` cluster — after backdating documents into old months and running the real retention drop:

```
dropped partitions: [documents_2026_02, documents_2026_03]
content_store rows 377 -> 377, bytes 11065936 -> 11065936 (UNCHANGED by the drop → orphaned blobs)
CONFIRMED: partition-drop retention orphans content_store blobs. Explicit DeleteDocument would reclaim them; retention does not.
```

## Impact

- Unbounded blob growth on the default Postgres backend — the single largest storage line item at scale.
- Silent: metadata dashboards look healthy (bounded, partitioned) while the blob table/bucket grows forever.
- Affects all backends (Postgres BYTEA / S3 / filesystem).

## Proposed design: split `content_store` by lifecycle

Replace the single shared `content_store` with two stores whose lifecycles match the data:

### 1. `document_content` — partitioned, retention-dropped (fixes the leak)

- Move document blobs into their own table (`document_content`), **RANGE-partitioned by `created_at`** with the documents retention window, registered as a `PartitionedTableContributor` alongside `documents`.
- Retention then reclaims blobs the same cheap way it reclaims metadata: `DROP TABLE <month>` — O(1) metadata, frees disk immediately, no per-row `DELETE`, no bloat, no `VACUUM`.
- **TTL vs partitioning:** a per-row `expires_at` TTL fixes *correctness* (a blob owns its own lifecycle, independent of the `documents` row) but not *efficiency* — bulk `DELETE … WHERE expires_at < now()` on large BYTEA/TOAST data leaves dead tuples that only `VACUUM (FULL)` reclaims, i.e. bloat + slow disk return, exactly what hurts at 100k/day. Partition-drop is the efficient path. With a **fixed** retention the **partition boundary *is* the TTL**, so no `expires_at` column is needed. Only add a per-row TTL column if we want **variable per-tenant/per-document** retention (which monthly partition-drop can't express at row granularity).

### 2. `asset_content` — content-addressable, ref-counted, keep-forever (dedup)

Assets repeat heavily (bundled fonts, system/demo-catalog images, the same logo across a tenant's templates), yet today identical bytes are stored once per `assetId`. Make asset storage **content-addressable**:

- Key `asset_content` by **content hash (SHA-256)** so identical bytes are stored once. The `assets` row becomes a pointer to a hash.
- **Deletion becomes ref-counted, not direct:** a blob may back many assets, so `DeleteAsset` can no longer delete the blob outright. Use a ref-count (increment on put-if-exists, decrement on delete, GC at zero) **or** a mark-and-sweep GC (drop `asset_content` hashes with no referencing `assets` row).
- **Dedup scope is a privacy decision — namespace the hash by scope `(system | <tenantId>, sha256)`:**
  - **System/bundled assets** (fonts, demo catalog) → dedup **globally**; high win, zero risk (shared by design).
  - **User-uploaded tenant assets** → dedup only **within the tenant**. Global cross-tenant dedup introduces an existence **side-channel** (a tenant can infer another uploaded the same file via put-if-exists timing) and complicates "delete all my data" / erasure (removing a reference leaves shared bytes). Per-tenant scope keeps most of the savings without the leak.

The `ContentStore` port stays a dumb KV; CAS key-derivation + ref-counting live in the **asset domain** above it, so all backends (Postgres/S3/filesystem) get dedup for free.

### 3. Backstop: orphan reaper + metric

- A `single_owner` scheduled reaper that deletes any blob with no live owner — catches drift and reclaims blobs **already** orphaned by past retention cycles (existing data). Scope it to the `documents/` set for reclamation and to zero-ref hashes for assets.
- Add a gauge (orphaned-blob count / bytes) so this can't silently regress.

## Migration

Data is stable (RC1+) — forward-only, no destructive rewrite. New timestamped migrations that:
1. Create `document_content` (partitioned) and `asset_content` (hash-keyed) tables.
2. Backfill: copy `documents/*` blobs → `document_content` (bucketed into the right monthly partitions by the owning document's `created_at`); copy `assets/*` blobs → `asset_content`, computing SHA-256 and deduplicating, and stamping `assets` rows with the hash.
3. Drop `content_store` once cutover is verified.

Backfill of large BYTEA data should be chunked/online.

## Alternatives considered

- **Cascade blob delete before partition drop** (enumerate a partition's doc ids and `contentStore.delete` each before `DROP`): fixes the leak without a new table, but couples partition maintenance to the content store and does nothing for asset dedup. Kept as a smaller-scope fallback if the split is deferred.
- **Blanket-partition/retain `content_store`**: rejected — it mixes keep-forever assets with droppable documents.
- **S3 bucket lifecycle rules**: S3-only and drifts from the app's retention config.

## Out of scope (separate issue)

`documents` / `document_generation_requests` list queries lack a `(tenant_key, created_at DESC)` index → deep-OFFSET pagination + the unbounded per-call `COUNT(*)` seq-scan across monthly partitions as a tenant's history grows.

---

_Found while building the `epistola-load-test` scale-verification tool (collect-balancing + volume/retention tracks); the retention-orphan and dedup behaviours were confirmed against a local `1.0.0-RC3` cluster._

sdegroot:
Fixes #738. Follow-ups tracked in #742 (teardown), #751 + contract #42 (sensitive-asset surfaces).

## Problem

Automatic partition-drop retention on `documents` never reclaimed the matching blobs in the shared, unpartitioned `content_store` — they orphaned and grew without bound (confirmed via `epistola-load-test`). `content_store` also mixed two lifecycles (ephemeral document PDFs vs keep-forever assets) and stored identical asset bytes once per asset.

## Solution — split blob storage by lifecycle

**Document content (ephemeral) → `DocumentContentStore`**, pluggable, each backend reclaiming natively:
- **PostgreSQL** (the only tested/supported backend): a RANGE-partitioned `document_content` table dropped **in lockstep** with the `documents` partitions — O(1) `DROP TABLE`, no per-row delete, no bloat. **This is the leak fix.**
- **S3 / filesystem** (**alpha, untested** — startup warning + docs): bucket lifecycle rule (toggleable, amends rather than clobbers) / filesystem age sweep.

**Asset content (permanent) → `AssetContentStore`**, always PostgreSQL, content-addressable (SHA-256), deduplicated:
- **Global dedup by default** — identical bytes (branding, images, fonts, the bundled system catalog) stored once installation-wide.
- **Sensitive assets** are stored isolated per-tenant (`sensitive ? tenant_key : 'global'`) — no cross-tenant existence side-channel, clean erasure. The backend is fully wired today; exposing the flag on UI/REST/MCP/catalog is #751 / contract #42.
- `DeleteAsset` no longer deletes bytes — the single-owner **`ContentReaper`** mark-and-sweeps unreferenced blobs and publishes `epistola.storage.orphaned_blobs`.

**Migration (transitional, grouped in `storage/backfill/` for clean removal in #742):**
- **`ContentBackfillRunner`** — a **background** single-owner cluster task (off the startup path, so it never races readiness on a large install), batched, resumable, idempotent, self-disabling via a completion marker. Moves legacy blobs forward (JVM-side hashing, so no `pgcrypto` dependency on the DDL-less runtime role). Safe during rolling upgrades — only upgraded nodes carrying the handler ever run it.
- **`LegacyBlobFallback`** — read paths fall back to `content_store` for a document not yet in `document_content` (closing a real multi-pod 404 gap) or an asset whose `content_hash` is still NULL.
- `content_store` is kept one release as a safety net, then dropped (#742).

## Notable

- **Backups ripple (R2):** `epistola-support-backups` read/wrote asset blobs from `content_store` by prefix — rewritten in lockstep to route through `asset_content` (guarded by round-trip tests), or restores would have silently broken.
- Breaking (`feat!`): the internal `ContentStore` port is split. No REST/MCP/UI contract change. Deliberate per the RC data-stability rules (data is preserved; only internal architecture changes).

## Tests

Document-retention lockstep reclaim; asset global dedup + sensitive per-tenant isolation; ref-aware delete + reaper GC + gauge; global system-asset dedup (the demo-catalog demonstration); filesystem age-sweep; S3 lifecycle merge (mocked); backfill idempotency + completion marker; legacy read-fallback (documents and assets); backup round-trip + cross-tenant isolation. Plus regression across assets/fonts/catalog/documents/storage, the backups module, cluster scheduling, and app-level asset/image/health handlers. Tests mint UUID-unique bytes so exact blob-count assertions can't collide in the shared DB. ktlint + oxfmt clean.

## Docs

New `docs/storage.md` (two-store architecture, per-backend retention, dedup/scope, reaper, backfill, backend support status, rolling-upgrade behaviour); `docs/cluster-resilience.md` §4 (scheduled tasks under version skew); `docs/metrics.md` (orphaned-blobs gauge); `docs/tenant-backup.md` (asset_content blob path).


modules/epistola-core/src/main/resources/db/migration/core/V20260714090000__core_document_content.sql
```diff
@@ -0,0 +1,36 @@
...
+-- backends reclaim their own way (bucket lifecycle rule / age sweep) and never
+-- touch this table. It is created for all backends for schema consistency, same
+-- as `content_store`.
+--
+-- The blob's `created_at` is the OWNING document's `created_at` (passed by the
+-- store), so the blob always lands in the same monthly partition as its metadata
+-- and ages out with it.
+
+CREATE TABLE document_content (
+    key          VARCHAR(512) NOT NULL,
```
sdegroot, line 21: since this always links to an actual document, why not add a foreign key to it?

modules/epistola-core/src/test/kotlin/app/epistola/suite/storage/AssetContentStoreIT.kt
```diff
@@ -0,0 +1,201 @@
...
+
+        val hashesA = systemAssetHashes(a.id)
+        val hashesB = systemAssetHashes(b.id)
+        assertThat(hashesA).isNotEmpty
+        assertThat(hashesB).containsExactlyInAnyOrderElementsOf(hashesA)
+        // Each bundled hash is stored exactly once in the global scope, regardless of tenant count.
+        hashesA.forEach { hash -> assertThat(blobRows(GLOBAL_ASSET_SCOPE, hash)).isEqualTo(1) }
+    }
+
+    private fun systemAssetHashes(tenant: TenantKey): List<String> = jdbi.withHandle<List<String>, Exception> { handle ->
```
sdegroot, line 160: I see a lot of direct sql in the tests. Ideally, all of our tests are without direct sql (unless no other way). Can we make this use more the commands we have (or create ones)

modules/epistola-core/src/main/resources/db/migration/core/V20260714090000__core_document_content.sql
sdegroot: Deliberately no FK, mainly because of **write ordering**: the blob is written *before* the `documents` row (`DocumentGenerationExecutor` — "store content first; an orphaned blob is harmless, a document with missing content is not"). An FK to `documents` would require the parent to exist first, forcing us to reverse that order and lose the crash-safety property (a crash between the two would then leave a document that 404s, instead of a harmless orphan blob the reaper cleans up).

Two supporting reasons: the `key` is the same `documents/{tenant}/{doc}` string the S3/filesystem backends use (uniform `ContentKey`), not the `documents (id, created_at)` PK — so an FK would need structured columns that diverge from the cross-backend key model; and both tables are RANGE-partitioned and dropped independently in lockstep by retention, so orphans are intentionally tolerated and reclaimed (partition-drop + reaper) rather than enforced by referential integrity.

Happy to revisit if you'd rather trade the write-first crash-safety for the FK guarantee, but I'd lean keep-as-is.

modules/epistola-core/src/test/kotlin/app/epistola/suite/storage/AssetContentStoreIT.kt
sdegroot: Good call — refactored (`92cd28d3`). `AssetContentStoreIT` now asserts blob presence through the `AssetContentStore.exists(scope, hash)` port and computes the expected hash from the uploaded bytes with `sha256Hex` (the same derivation the write path uses), instead of `SELECT`ing `asset_content`/`assets`. Since `(scope, content_hash)` is the table PK, `exists` is exactly "one blob".

Two SQL helpers remain, both commented as the "no other way" exception: reading the bundled **system-catalog** assets' `content_hash` (no command creates them in the test and no query surfaces the hash), and planting a historical `created_at` for the reaper grace-window test (no command sets `asset_content.created_at`). The other new storage/backfill tests keep raw SQL only for seeding legacy/transitional state that no command produces (e.g. inserting a pre-migration `content_hash IS NULL` asset, writing directly to the soon-removed `content_store`, or creating an ancient partition).

sdegroot: Correction to my earlier reply — I'd only refactored `AssetContentStoreIT` at that point, not the others. I've now audited every new/changed test in the PR and given them the same treatment (`75a7fc33`): `ContentBackfillRunnerIT`, `BackupAssetBlobIntegrationTest`, and `CrossTenantIsolationIntegrationTest` now check/read `asset_content` through the `AssetContentStore` port and derive hashes with `sha256Hex` instead of `SELECT`ing `asset_content` / `assets.content_hash`.

What raw SQL remains is either **seeding legacy/transitional state no command produces** (writing to `content_store`, inserting a pre-migration `content_hash IS NULL` asset, creating an ancient partition, clearing the backfill marker) or a **compact multi-table row-count probe** (the isolation test's "did anything change?" check) — each now has a one-line comment saying why.

PR #765, opened 2026-07-22
feat(editor): guided editor walkthrough (alpha)

whit3st:
## Description

Adds a guided, in-editor walkthrough of the template editor (driver.js), behind the off-by-default **alpha** `editor-walkthrough` feature toggle (`epistola.features.editor-walkthrough`), resolved per tenant.

- Organised as short, independently-editable **chapters** driven one at a time, with per-chapter progress; finishing a chapter offers "Next: …" to continue.
- A toolbar **Guide** button opens a chapter menu to start or replay any chapter. The tour never auto-starts; on a user's first visit a one-time coach-mark points at the Guide button.
- When the flag is off, **none of it is downloaded** — the launcher, tour code, and driver.js all sit behind gated dynamic imports; the always-loaded bundle is unchanged.
- Threads the tenant toggle into the editor host as an `EditorFeatureFlags` flag (the editor's first use of that extension point).

Purely a client-side onboarding overlay — no data is read or sent — so it has no REST/MCP surface and no catalog artifact.

## Related Issue(s)

None.

## Type of Change

## Component(s) Affected

- [x] Backend (Spring Boot/Kotlin) — feature-toggle registration and editor-handler wiring
- [x] Frontend (Editor/TypeScript) — walkthrough runner, launcher, tours

## Checklist


modules/editor/src/main/typescript/ui/EpistolaInspector.ts
```diff
@@ -162,7 +162,7 @@ export class EpistolaInspector extends LitElement {
     const docStyles = (this.doc?.documentStylesOverride ?? {}) as Record<string, unknown>;
 
     return html`
-      <div class="inspector-section">
+      <div class="inspector-section" data-tour="document-styles">
         <div class="inspector-section-label">Document Styles</div>
```
whit3st, line 174: i wonder if we can make these data-testid so that we can use them also in playwright tests

docs/feature-toggles.md
```diff
@@ -135,10 +135,11 @@ rule in `components.css`.
...
-| `support-feedback`            | Support → Feedback (local; hub sync gated on the support tier)       | Stable | `true`  |
-| `support-backups`             | Support → Backups (faithful full-fidelity tenant backups + restore)  | Beta   | tier\*  |
-| `support-compatibility-check` | Support → Upgrading (compatibility checks against upcoming releases) | Stable | tier\*  |
+| Key                           | Description                                                           | Stage  | Default |
+| ----------------------------- | --------------------------------------------------------------------- | ------ | ------- |
+| `support-feedback`            | Support → Feedback (local; hub sync gated on the support tier)        | Stable | `true`  |
+| `support-backups`             | Support → Backups (faithful full-fidelity tenant backups + restore)   | Beta   | tier\*  |
+| `support-compatibility-check` | Support → Upgrading (compatibility checks against upcoming releases)  | Stable | tier\*  |
+| `editor-walkthrough`          | Guided in-editor tour of the template editor (client-side onboarding) | Alpha  | `false` |
```
whit3st, line 146: hmm quality is missing here, out of scope but something to take care of

modules/editor/src/main/typescript/ui/walkthrough/launcher.ts
```diff
@@ -0,0 +1,339 @@
...
+.ep-wt-mark {
+  flex: 0 0 auto; display: inline-flex; align-items: center; justify-content: center;
+  width: 1.25rem; height: 1.4rem; color: var(--ep-stone-400);
+}
+.ep-wt-mark.is-done { color: var(--ep-success); }
+.ep-wt-item.is-current .ep-wt-mark { color: var(--ep-primary-strong); }
+.ep-wt-text { display: flex; flex-direction: column; gap: var(--ep-space-0-5); }
+.ep-wt-name { font-weight: 500; color: var(--ep-stone-800); }
+.ep-wt-summary { font-size: var(--ep-text-xs); color: var(--ep-stone-600); line-height: 1.4; }
+`;
```
whit3st: for better readability and consistency, should this styling be in its own css file?

modules/editor/src/main/typescript/ui/walkthrough/launcher.ts
```diff
@@ -0,0 +1,339 @@
...
+  private _close(refocusTrigger = true): void {
+    if (!this._open) return;
+    this._open = false;
+    if (refocusTrigger) {
+      this.querySelector<HTMLButtonElement>('[data-testid="walkthrough-guide-trigger"]')?.focus();
+    }
+  }
+
+  private _items(): HTMLButtonElement[] {
+    return Array.from(this.querySelectorAll<HTMLButtonElement>('.ep-wt-item'));
```
whit3st: bunch of hard-coded selector strings. they could be exported to a file for shared usage and better debugging/readibility?

modules/editor/src/main/typescript/ui/walkthrough/walkthrough.ts
```diff
@@ -0,0 +1,267 @@
...
+.driver-popover-next-btn:hover { background: var(--ep-primary-strong, #9d5741); }
+.driver-popover-close-btn { color: var(--ep-stone-400, #b9a595); }
+.driver-popover-close-btn:hover { color: var(--ep-stone-800, #46302b); }
+/* The rich-text bubble menu sits at --ep-z-dropdown (10), far below driver's
+   overlay (10000), so it renders *behind* the backdrop when a text editor is
+   spotlighted. While a tour is active, lift it above the backdrop (but below the
+   popover) and keep it interactive so it reads as a real floating toolbar. */
+.driver-active .pm-bubble-menu { z-index: 1000000; }
+.driver-active .pm-bubble-menu, .driver-active .pm-bubble-menu * { pointer-events: auto; }
+`;
```
whit3st: these also can be in a css file, same file that above comment mentions

modules/editor/src/main/typescript/plugins/walkthrough/walkthrough.ts
```diff
@@ -0,0 +1,267 @@
...
+function ensureDriverStyles(): void {
+  injectStyleOnce(STYLE_ID, driverCss + POPOVER_OVERRIDES);
+}
+
+/**
+ * Resolve a step's target within this editor host (not the whole document), so
+ * spotlights land on the right editor when more than one is mounted. Returns
+ * `null` when absent; driver's `skipMissingElement` then skips the step.
+ */
+function hostTarget(host: HTMLElement, selector: string): () => Element {
```
whit3st, line 52: find out if these 'selector' parameters are known. if so, making this 'string' type stronger is a better option. maybe string literal unoin? maybe not a good idea, check

modules/editor/src/main/typescript/ui/EpistolaInspector.ts
whit3st: also we declare and export these variables already in TOUR_HOOKS variable...

whit3st: Done in 3c919dad — components now bind `data-tour=${TOUR_HOOKS.x}`, so renames are compile-checked. Moved the module to `ui/tour-hooks.ts` (components shouldn't import from the walkthrough dir; it stays dependency-free). The test now matches key references — it still exists because the compiler only proves both sides agree on the name, not that the attribute is actually stamped.

modules/editor/src/main/typescript/ui/walkthrough/launcher.ts
whit3st: Extracted to `launcher.css` in 10bf07fe, imported with `?inline` (same mechanism as driver.css) so it keeps riding the lazy chunk instead of editor.css. Also now covered by `pnpm lint:css`.

modules/editor/src/main/typescript/ui/walkthrough/walkthrough.ts
whit3st: Same commit (10bf07fe) — the popover re-skin is now `walkthrough.css`, `?inline`, concatenated with driver's CSS exactly as before.

modules/editor/src/main/typescript/ui/walkthrough/launcher.ts
whit3st: Named them as module-local consts in ad08bf02 (and folded the duplicated `closest('epistola-editor')` into `_hostEl`). Kept them private to the launcher rather than a shared file — nothing else should query its internals.

modules/editor/src/main/typescript/plugins/walkthrough/walkthrough.ts
whit3st: Done across two commits. 38b4b65d closes the grammar: `TourStep.target` is now a template-literal union — a `[data-tour~="<hook>"]` selector or a registered epistola element (derived from `HTMLElementTagNameMap`, so a typo'd tag is a compile error with a did-you-mean). Internal class/id selectors no longer typecheck; the `data-testid` arm was dropped by converting the two remaining testid targets to hooks. 481e3438 then removes `tourHook()` entirely — with the type doing the checking the builder was redundant sugar, so targets are written as typed literals (the dynamic `tab-${id}` case stays checked via template-literal distribution). Verified with in-situ typo probes and a full clean build incl. uiTest.

modules/editor/src/main/typescript/ui/EpistolaInspector.ts
```diff
@@ -127,7 +128,7 @@ export class EpistolaInspector extends LitElement {
         <!-- Delete -->
         ${!presentation?.suppressDeleteSection
           ? html`
-              <div class="inspector-delete-section">
+              <div class="inspector-delete-section" data-tour=${TOUR_HOOKS.blockDelete}>
```
sdegroot: do we need to put the tour in the editor like this? Now its a bidirection awareness.  The tour should be aware of all the components etc. but not vice versa. Its fine if we need to add some unique id's or something but couldn't we use testids for this?

modules/editor/src/main/typescript/plugins/walkthrough/launcher.css
```diff
@@ -0,0 +1,116 @@
+/*
```
sdegroot, line 1: Could we make the tour a separate plugin maybe? instead of having it in the main code?

modules/editor/src/main/typescript/ui/EpistolaInspector.ts
whit3st: what the component imports is just tour-hooks.ts which is a dependency-free constants file, no tour code loads from it. so the component isnt really aware of the tour, its more like declaring a stable anchor, basically the unique id you mention but typed, so renames are compile checked and verified by test. these were string literal before, i wanted to do this deliberately. let me know what you think

modules/editor/src/main/typescript/plugins/walkthrough/launcher.css
whit3st: yeah i thought about it but the plugin api needs extending first e.g. plugins dont get the host element that tours scope their spotlights to so i kept it out of scope but i think this can be a good follow-up refactor

modules/editor/src/main/typescript/ui/EpistolaInspector.ts
sdegroot: Agreed — changed this in `1ffa8bd73`. Editor components now expose generic `data-editor-anchor` values from `EDITOR_UI_ANCHORS`; there is no walkthrough/tour import or naming on the editor side. The walkthrough owns the selector grammar that consumes those anchors. I kept this separate from `data-testid` because these are stable integration anchors for editor plugins, while test IDs remain a testing-only contract. The bindings are covered by regression tests.

modules/editor/src/main/typescript/plugins/walkthrough/launcher.css
sdegroot: Done in `0f19c1a8f` and `6e288a410`. The walkthrough is now a separate frontend plugin and Vite entry, loaded through the generic editor plugin descriptor only when the feature is enabled. The generic API gained unified `toolbarItems` plus the current `PluginContext`; the main toolbar no longer contains walkthrough-specific behavior, and the walkthrough/driver assets remain lazy-loaded.

PR #771, opened 2026-07-22
feat(quality): add editor quality panel

sdegroot:
## Summary

- adds the quality source input requirement model, including resolved template dependencies for checks that need extra draft context
- keeps the quality ADRs as draft discussion records and adds a draft ADR for typed editor plugin selection intents
- adds internal editor quality routes for variant findings and explicit check-now
- wires the template editor to show a feature-gated Quality sidebar panel with stale markers, check-now, refresh, and node navigation
- keeps Quality open only for finding-driven navigation, while normal manual selections still move to Inspector
- improves sidebar tab overflow so plugin tabs remain discoverable without a persistent scrollbar
- removes the alpha-stage daily quality sweep; in-process checks now run on publish and explicit editor Check now, with a future reconciliation sweep documented as planned once run lifecycle/source tiering exists

## Validation

- `./gradlew :modules:epistola-quality:unitTest :modules:epistola-quality:ktlintCheck`
- `./gradlew :modules:epistola-quality:integrationTest`
- `./gradlew :apps:epistola:test --tests app.epistola.suite.quality.QualityHandlerHtmxTest`
- `./gradlew :modules:epistola-quality:ktlintCheck :apps:epistola:ktlintCheck`
- `./gradlew :modules:epistola-quality:test`
- `pnpm --filter @epistola/editor test -- src/main/typescript/ui/EpistolaSidebar.test.ts src/main/typescript/ui/quality-plugin.test.ts src/main/typescript/plugins/plugins.test.ts src/main/typescript/ui/EpistolaEditor.test.ts`
- `pnpm --filter @epistola/editor build`
- `pnpm format:check`
- `git diff --check`


apps/epistola/src/main/resources/templates/templates/editor.html
```diff
@@ -92,7 +92,12 @@
         "variantId": [[${variantId}]],
         "dataExamples": [[${dataExamples}]],
         "dataModel": [[${dataModel}]],
-        "locale": [[${resolvedLocale}]]
+        "locale": [[${resolvedLocale}]],
+        "quality": {
```
sdegroot, line 97: why not use the feature toggle feature for htis?

modules/editor/src/main/typescript/plugins/types.ts
```diff
@@ -23,6 +23,9 @@ export interface PluginContext {
 
   /** Currently selected node, or null */
   selectedNodeId: NodeId | null;
+
+  /** Persist the latest document before a plugin asks the host for draft-bound work. */
```
sdegroot: why do we add this function in this PR? what does it have to do with quality checks?

apps/epistola/src/main/resources/templates/templates/editor.html
sdegroot: Yes, agreed. This now uses the existing tenant feature-toggle path: `DocumentTemplateHandler` resolves `ResolveFeatureToggles`, passes `featureFlags.quality` into the editor config, and `editor-boot.js` only registers the Quality plugin when `config.featureFlags?.quality` is true. The old dedicated `qualityEnabled` style flag is gone.

modules/editor/src/main/typescript/plugins/types.ts
sdegroot: This is part of the Quality panel behavior rather than a generic unrelated plugin addition. Selecting a Quality finding needs to select/reveal the referenced block while keeping the Quality panel open for that plugin-driven navigation only. I replaced the earlier tab-level `keepOpenOnSelection` metadata with this typed host action so the behavior is attached to the selection intent, not to the tab. That avoids the bug where manual canvas selections while Quality is open would also keep the Quality tab active. I also added draft ADR 0013 to document this plugin-to-host UI intent pattern.

PR #779, opened 2026-07-24
fix(api): harden template document graph handling

sdegroot:
## What changed

This hardens Epistola's template document graph handling around the REST/API paths that the fuzz suite exercised.

- Adds strict template graph validation, including synthetic `root` requirements, cycle detection, reachability checks, size/depth limits, and supported node type checks.
- Wires graph validation into draft, version, stencil, and import flows so invalid content is rejected before persistence or rendering.
- Adds structured API handling for draft lifecycle conflicts and row-level locking around draft/version lifecycle operations.
- Makes template path extraction cycle-safe.
- Updates built-in catalog fixtures for the stricter root graph contract and bumps the demo catalog fixture version.
- Adds regression and integration coverage for invalid graphs, unsupported node types, draft retention, and concurrent lifecycle behavior.

## Why

The fuzz testing work found that malformed template document graphs could trigger incorrect behavior or server errors instead of controlled validation failures. The root cause was that template graph shape was assumed by downstream render/path code but was not enforced consistently at the API boundary.

## Impact

Templates must now use exactly one synthetic document root node with type `root`. Invalid, cyclic, unreachable, oversized, or unsupported graphs fail validation with structured errors instead of being retained or passed to rendering.

## Validation

- `./gradlew ktlintFormat ktlintCheck unitTest integrationTest` was run; the full run surfaced only a stale demo catalog version expectation after the fixture bump.
- `./gradlew ktlintFormat :modules:epistola-core:integrationTest` passed after updating that expectation.
- Earlier focused API/fuzzer regressions passed during the fix cycle.


modules/epistola-core/src/main/kotlin/app/epistola/suite/catalog/commands/ImportStencil.kt
```diff
@@ -76,9 +77,11 @@ data class ImportStencilResult(
 class ImportStencilHandler(
     private val jdbi: Jdbi,
     private val objectMapper: ObjectMapper,
+    private val graphValidator: TemplateDocumentGraphValidator,
 ) : CommandHandler<ImportStencil, ImportStencilResult> {
 
     override fun handle(command: ImportStencil): ImportStencilResult {
+        graphValidator.validate(command.content)
```
sdegroot: the content of a stencil is not necessarily starting with a root, should it?

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/validation/TemplateDocumentGraphValidator.kt
```diff
@@ -0,0 +1,174 @@
...
+import app.epistola.suite.validation.ValidationException
+import app.epistola.template.model.TemplateDocument
+import org.springframework.stereotype.Component
+
+/**
+ * Validates the node/slot graph invariants shared by templates and stencil
+ * definitions before deeper validators, path extraction, or rendering traverse
+ * the document.
+ */
+@Component
```
sdegroot: does this need to be a spring component?

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/analysis/TemplatePathExtractor.kt
```diff
@@ -45,7 +45,9 @@ class TemplatePathExtractor {
         document: TemplateDocument,
         scopeStack: List<LoopScope>,
         paths: MutableSet<String>,
+        visiting: MutableSet<String>,
```
sdegroot: what exactly is happening here?

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/validation/NodeParameterBindingValidator.kt
```diff
@@ -37,7 +37,7 @@ class NodeParameterBindingValidator(
      * snapshot) — those flow through with whatever schema-shaped bindings were present.
      */
     @Suppress("UNCHECKED_CAST")
-    fun validate(doc: TemplateDocument) {
+    fun validate(doc: TemplateDocument, fieldPrefix: String = "content") {
```
sdegroot: why do we need this fieldPrefix?

modules/epistola-core/src/main/kotlin/app/epistola/suite/catalog/commands/ImportStencil.kt
sdegroot: Yes. We decided that every stored stencil is a complete `TemplateDocument`, not a fragment, and therefore has exactly one `root` node. Stencil insertion treats that root as a structural envelope and embeds its children. Supporting rootless stencil content would require a separate fragment model and parallel traversal rules. Existing editor-created and catalog stencils already use this shape; malformed payloads created only through the previously unvalidated REST path are intentionally not supported.

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/validation/TemplateDocumentGraphValidator.kt
sdegroot: Agreed; it did not need to be one. I removed `@Component` and made `TemplateDocumentGraphValidator` an internal, pure implementation owned by the `TemplateDocumentValidator` facade. `GenerationService` now calls the facade through the explicit graph-only `validateTemplateGraphForRendering` method, which preserves database compatibility without exposing the lower-level validator as application policy.

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/analysis/TemplatePathExtractor.kt
sdegroot: This is a defensive DFS cycle guard. `activePath` contains only node IDs on the current recursion path; seeing the same ID again stops traversal before a malformed legacy or imported graph can cause `StackOverflowError`. Graph validation remains authoritative and runs first on normal write paths. I renamed `visiting` to `activePath` and added a cyclic-document regression test to make that behavior explicit.

modules/epistola-core/src/main/kotlin/app/epistola/suite/templates/validation/NodeParameterBindingValidator.kt
sdegroot: The prefix existed only to map document-relative errors onto the request field: template payloads use `templateModel`, while stencil payloads use `content`. Passing an arbitrary string through every low-level validator was the wrong abstraction. I removed `fieldPrefix` from those validators; they now report exact relative paths such as `nodes.stencil-1.props.parameterBindings.param1`, and only the `TemplateDocumentValidator` facade adds `templateModel` or `content`.

PR #800, opened 2026-07-30
feat(ui): wider content area with fixed list table layouts

whit3st:
## Description

Widens the regular-page content area from 56rem to 72rem and converts all list tables to fixed column layouts so they provably never overflow the page, no matter how hostile the user data.

- `main.regular` / `.main-content` now use `--ep-max-width-lg` (72rem).
- New `.ep-table-fixed` modifier in the design system: `table-layout: fixed` with percentage `<colgroup>`s (totaling 100%, so browsers normalize and the table can never exceed its container), single-line cells that clip with an ellipsis, and a `cell-wrap` escape hatch for badge/tag lists (where a single overlong badge ellipsizes inside its own pill).
- Applied to all ten list tables: templates, stencils, themes, tenants, catalogs, catalog browsing, code lists, attributes, environments, API keys.
- Every truncatable cell (names, slugs, IDs, descriptions, dates, catalog keys) carries a `th:title` so hover reveals the full value; null values correctly omit the attribute.
- Status badges that must never be hidden ("Default", "Pin drift", "refresh failed") are placed *before* the variable-length text in their cells, so a long name truncates instead of pushing the badge out of view.
- Permission-gated Actions columns (environments, attributes) gate their `<col>`/`<th>` on the same permission as the `<td>`, so viewer roles don't get a blank reserved column under fixed layout.

Verified in a live browser at multiple viewports (measured cell geometry via DOM, not just visual inspection), including hostile long-name demo data and both admin and viewer roles.

## Related Issue(s)

—

## Type of Change

- [ ] Documentation update

## Component(s) Affected

- [ ] Frontend (Editor/TypeScript)
- [ ] Documentation
- [ ] CI/CD

Thymeleaf templates and CSS only — no Kotlin changes.

## Checklist

No new automated tests: the change is purely visual layout (column widths, clipping, hover titles), which the existing `CspTemplateComplianceTest` and per-page `*HandlerHtmxTest`s already cover structurally; layout geometry itself was verified by live browser measurement. `unitTest` + `integrationTest` pass locally on the rebased branch; the Playwright `uiTest` suite is left to CI.

## Screenshots (if applicable)

Not attached; happy to add before/after captures on request. The visible change: pages are wider, and an extremely long name/slug/ID now truncates with an ellipsis (full value on hover) instead of stretching the table off the page.

## Additional Notes

- The percentage-colgroup approach is documented in `modules/design-system/components.css` next to the `.ep-table-fixed` rule, including why large pixel `<col>` widths must be avoided and when to use `cell-wrap`.
- Badge-before-text placement in the themes/browse/code-lists cells is deliberate and commented in the templates — the cell clips at its end, so a trailing badge would be pushed out of view by a long name.
- Tables outside the ten list pages (feedback, backups, quality, audit, logs, fonts) are intentionally not converted in this PR.

### DIFF
--- a/apps/epistola/src/main/resources/templates/api-keys/list.html
+++ b/apps/epistola/src/main/resources/templates/api-keys/list.html
@@ -36,7 +36,17 @@
             </button>
         </div>
 
-        <table th:if="${!#lists.isEmpty(apiKeys)}" class="ep-table">
+        <table th:if="${!#lists.isEmpty(apiKeys)}" class="ep-table ep-table-fixed">
+            <colgroup>
+                <col style="width: 13%" /><!-- Name -->
+                <col style="width: 13%" /><!-- Prefix -->
+                <col style="width: 18%" /><!-- Scope -->
+                <col style="width: 9%" /><!-- Created by -->
+                <col style="width: 15%" /><!-- Created -->
+                <col style="width: 15%" /><!-- Last used -->
+                <col style="width: 11%" /><!-- Expires -->
+                <col style="width: 6%" /><!-- Actions -->
+            </colgroup>
             <thead>
                 <tr>
                     <th>Name</th>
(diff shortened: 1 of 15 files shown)


apps/epistola/src/main/resources/templates/api-keys/list.html
```diff
@@ -36,7 +36,17 @@
             </button>
         </div>
 
-        <table th:if="${!#lists.isEmpty(apiKeys)}" class="ep-table">
+        <table th:if="${!#lists.isEmpty(apiKeys)}" class="ep-table ep-table-fixed">
+            <colgroup>
```
sdegroot, line 43: previously this was auto generated. now not anymore, what are the consequences?

apps/epistola/src/main/resources/templates/api-keys/list.html
```diff
@@ -52,28 +62,28 @@
...
                             </span>
                         </td>
-                        <td>
+                        <td th:title="${apiKey.createdByDisplayName}">
                             <span th:if="${apiKey.createdByDisplayName != null}" th:text="${apiKey.createdByDisplayName}"></span>
                             <span th:if="${apiKey.createdByDisplayName == null}" class="text-muted">—</span>
                         </td>
-                        <td th:text="${#temporals.format(apiKey.createdAt, 'yyyy-MM-dd HH:mm')}"></td>
-                        <td>
+                        <td th:text="${#temporals.format(apiKey.createdAt, 'yyyy-MM-dd HH:mm')}" th:title="${#temporals.format(apiKey.createdAt, 'yyyy-MM-dd HH:mm')}"></td>
```
sdegroot, line 83: nothing to do with this PR but i wonder: is this localized in the browsers timezone?

apps/epistola/src/main/resources/templates/attributes/list.html
```diff
@@ -45,24 +45,34 @@
             </button>
         </div>
 
-        <table th:if="${!#lists.isEmpty(attributes)}" class="ep-table">
+        <table th:if="${!#lists.isEmpty(attributes)}" class="ep-table ep-table-fixed">
+            <colgroup>
+                <col style="width: 19%" /><!-- ID -->
```
sdegroot, line 53: same question as the other table

apps/epistola/src/main/resources/templates/catalogs/browse.html
```diff
@@ -26,7 +26,18 @@ <h2>Resources</h2>
                         data-open-dialog="install-preview-dialog">Install All</button>
             </div>
             <div class="card-content">
-                <table class="ep-table">
+                <table class="ep-table ep-table-fixed">
+                    <!--/* Authored catalogs omit the last two cols; the sub-100% total
+                           is fine — browsers scale percentage colgroups to fit. */-->
+                    <colgroup>
```
sdegroot, line 34: same

apps/epistola/src/main/resources/templates/api-keys/list.html
sdegroot: Correct — the browser previously derived the column widths from content. With `table-layout: fixed`, the `<colgroup>` becomes a maintained layout contract: hostile values can no longer stretch the page, but widths are less adaptive and must stay aligned with the rendered columns. I’ve added a shared keyboard-focusable horizontal scroller for narrow viewports, given the table a desktop-width minimum, compacted/wrapped Actions cells, and added a 640px Playwright regression that verifies the page does not overflow and the delete action remains fully visible and clickable.

apps/epistola/src/main/resources/templates/api-keys/list.html
sdegroot: No — this timestamp is currently rendered server-side by Thymeleaf, not localized in the browser. For these `Instant` values Thymeleaf uses the JVM default timezone; `OffsetDateTime` values elsewhere retain their stored offset. This PR does not change that behavior. Browser-local rendering already exists for logs/audit/cluster and should be handled as a separate follow-up if we want it consistently across list pages.

apps/epistola/src/main/resources/templates/attributes/list.html
sdegroot: Same tradeoff: the `<colgroup>` is now an explicit layout contract rather than content-derived sizing. The permission gating here remains synchronized—the Actions `<col>`, `<th>`, and `<td>` are all omitted for viewers. The shared responsive scroller/minimum width now keeps the fixed columns usable on narrow screens, and long allowed-value badges now expose their full value via `title`.

apps/epistola/src/main/resources/templates/catalogs/browse.html
sdegroot: Same answer here: the widths are explicit and maintained with the rendered columns. Authored catalogs correctly omit Status and Actions, and the remaining columns expand to fill the table. The shared responsive scroller/minimum width now contains the table on narrow screens without clipping its content or actions.

PR #807, opened 2026-08-01
chore(toolchain): adopt TypeScript 7 (Go port) and pnpm 11

whit3st:
## Summary

Adopts the TypeScript 7 Go port and pnpm 11 for the frontend toolchain. Supersedes the parked `experiment/typescript-7-go-port` branch (136 commits behind; redone fresh rather than rebased).

- **TypeScript 6.0.3 → 7.0.2**, pinned exactly at root and in `modules/editor`. Cold `tsc -b`: ~3.9 s → ~0.45 s (~8×), byte-identical diagnostics, zero source/tsconfig changes.
- **pnpm 10.33.2 → 11.18.0** in both `.mise.toml` and `packageManager` — no self-delegation split. Requires mise ≥ 2026.7 locally (`mise self-update` if `mise install` reports `no asset found: pnpm-linux-x64`); CI's unpinned `mise-action` is unaffected.
- **Supply-chain gate**: `minimumReleaseAge: 10080` (7 days), with `@epistola.app/**` exempt — contract packages are exact-pinned in lockstep with the backend (`checkContractVersionAlignment`), so no older version can satisfy the range. Renovate carries a matching `minimumReleaseAge: "7 days"`, and `typescript` joins the non-automerged `toolchain` group.
- **pnpm 11 fallout handled**: `onlyBuiltDependencies` → `allowBuilds` (+ explicit `vue-demi: false`); the new builtin `sbom` shadows the editor package script, so CI/docs use `pnpm --filter @epistola/editor run sbom`; the dead root `clean` script (same collision) is removed; `.pnpm-store/` is now gitignored.

## Verification

- Diagnostics parity: `tsc -b --force` output identical under 6.0.3 and 7.0.2 (zero diagnostics both)
- 2004 editor tests; full `./gradlew build` (2917 backend tests, zero failures)
- `checkContractVersionAlignment` passes — pnpm 11 keeps lockfile format 9.0
- REUSE lint, oxlint/oxfmt, SBOM + notices generation
- CI `build` job run end-to-end locally with act: cold-container mise → pnpm 11 fetch, `--frozen-lockfile` under the supply-chain policy, SBOM generation via the fixed command, Trivy scans clean. Only the three `actions/upload-artifact` steps fail under act (its bundled artifact server predates the action's `mime_type` protocol field — act-only, does not affect GitHub runners)
- Editor SBOM diffed against the v1.0.0 release artifact: same generator (cdxgen 12.8.2) and spec (CycloneDX 1.6); component diff is exactly typescript 6→7, the contract 1.0.0 bump already on main, and one release-age downgrade

## Reviewer notes

- Four lockfile entries are one release older than on main (`conventional-commits-parser`, `globals`, `lucide-static`, `yargs`): they were younger than the 7-day gate at resolution time. They age past the cutoff within ~36 h of this PR and Renovate re-proposes them — no action needed.
- TypeScript minors ship breaking type changes, hence the exact pin; future TS releases arrive as reviewable toolchain-group PRs instead of automerges.

## Follow-ups (deliberately out of scope)

- `oxlint-tsgolint` re-versioned to track TypeScript (`^0.24.0` can never match the new `7.x` scheme); oxlint 1.75 already wants `>=7.0.2001`. Aligning it changes lint output independently of this upgrade.
- The SBOM script fetches cdxgen unpinned via `npx --yes` — outside the new release-age gate; worth pinning or moving into devDependencies.
- Renovate flags `gradle.fileMatch` as renamed to `managerFilePatterns` (pre-existing; old name still works).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,8 +59,3 @@ plan.md
 # workspace as a Docker volume). Never commit store blobs; reuse lint skips
 # gitignored files.
 .pnpm-store/
-
-### Claude Code ###
-# Per-developer local settings (permission allowlists); the shared
-# .claude/settings.json stays tracked.
-.claude/settings.local.json
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,15 @@ dist/
 ### Test Coverage ###
 coverage/
 plan.md
+
+### pnpm ###
+# pnpm falls back to an in-project store when the repo and the global store
+# live on different filesystems (also happens under act, which mounts the
+# workspace as a Docker volume). Never commit store blobs; reuse lint skips
+# gitignored files.
+.pnpm-store/
+
+### Claude Code ###
+# Per-developer local settings (permission allowlists); the shared
+# .claude/settings.json stays tracked.
+.claude/settings.local.json
--- a/package.json
+++ b/package.json
@@ -4,8 +4,8 @@
   "description": "Epistola Suite - Document generation platform",
   "scripts": {
     "build": "pnpm -r build",
+    "typecheck": "tsc -b modules/editor",
     "dev": "pnpm -r --parallel dev",
-    "clean": "pnpm -r clean",
     "test": "pnpm -r test",
     "test:watch": "pnpm -r test:watch",
     "test:coverage": "pnpm -r test:coverage",
@@ -30,10 +30,10 @@
     "oxlint-tsgolint": "^0.24.0",
     "stylelint": "^17.11.0",
     "stylelint-config-standard": "^40.0.0",
-    "typescript": "^6.0.0"
+    "typescript": "7.0.2"
   },
   "engines": {
     "node": ">=24.0.0"
   },
-  "packageManager": "pnpm@10.33.2"
+  "packageManager": "pnpm@11.18.0"
 }
--- a/renovate.json
+++ b/renovate.json
@@ -5,7 +5,8 @@
   "prHourlyLimit": 5,
   "prConcurrentLimit": 10,
   "ignorePaths": ["**/test/resources/**"],
-  "constraints": { "pnpm": "10.28.2", "node": "24.18.0" },
+  "constraints": { "pnpm": "11.18.0", "node": "24.18.0" },
+  "minimumReleaseAge": "7 days",
   "lockFileMaintenance": {
     "enabled": true,
     "groupName": "all non-major dependencies",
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -4,8 +4,8 @@
   "description": "Epistola Suite - Document generation platform",
   "scripts": {
     "build": "pnpm -r build",
+    "typecheck": "tsc -b modules/editor",
```
whit3st, line 7: why add this out of the blue?

renovate.json
```diff
@@ -34,15 +35,16 @@
...
         "org.jetbrains.kotlin*",
         "org.jetbrains.kotlinx.kover*",
         "org.jetbrains.kotlinx:kover-gradle-plugin",
         "gradle",
         "java",
         "node",
-        "pnpm"
+        "pnpm",
+        "typescript"
       ],
```
whit3st, line 48: why add ts here?

.gitignore
```diff
@@ -52,3 +52,15 @@ dist/
...
+# pnpm falls back to an in-project store when the repo and the global store
+# live on different filesystems (also happens under act, which mounts the
+# workspace as a Docker volume). Never commit store blobs; reuse lint skips
+# gitignored files.
+.pnpm-store/
+
+### Claude Code ###
+# Per-developer local settings (permission allowlists); the shared
+# .claude/settings.json stays tracked.
+.claude/settings.local.json
```
whit3st: what? is this related to this update branch?

package.json
whit3st: convenience, wouldnt hurt i guess

whit3st: Standalone type-check entry point. `pnpm build` only type-checks as a side effect of the full editor build (tsc + vite + registry dump/check); `pnpm typecheck` runs just `tsc -b`, which finishes in ~0.5 s under TS 7 — quick enough for a pre-commit sanity check.

renovate.json
whit3st: typescript is pinned exactly because its minor releases ship breaking type changes — the same reason Kotlin, Gradle, Node, and pnpm live in this group. Without this entry, TS updates automerge silently inside the "all non-major dependencies" group; with it, every TS release arrives as a non-automerged toolchain PR (titled with the exact version when it is the only update in the batch), so compiler upgrades stay deliberate and visible.

.gitignore
whit3st: Removed in da46d461a — agreed, not related to this branch. Real CI clones the repo and never sees untracked local files; the entry only mattered for local act runs (which copy the working directory).

Issue #477, opened 2026-05-27
[UI]: no consistent error/success feedback for async actions across all pages


whit3st:
## Description

The frontend has ~80 async HTTP requests and every one handles errors and success feedback differently, or not at all. There is no shared pattern, utility, or design convention.

### Current patterns

| Pattern | Count | Where |
|---|---|---|
| **HTMX global handler only** | ~50 | Most `hx-*` calls use generic "An unexpected error occurred" banner at the top of the page |
| **`fetch()` without `.catch()`** | 5 templates, 13 editor calls | Network errors silently swallowed, user sees nothing |
| **`alert()`** | 2 (`pdf-preview.js`, `version-comparison.html`) | Intrusive, unstyled, not accessible |
| **Custom HTMX error listeners** | 4 (font/asset uploads, catalog create/install) | Each has its own inline handling: different markup, no reuse |
| **Structured error return** | Data contract editor, theme save | Best pattern, but editor-bound only |

### Key gaps

- **PDF/A toggle, theme selector, template rename** (`settings.html`): the fetch calls lack `.catch()`. A network error produces zero feedback beyond a silently-reverted checkbox.
- **Editor calls** (`editor.html` ~15 fetch calls): most parse errors and throw, but callers (`SaveService`, `PreviewService`) silently swallow them. User never sees the error.
- **Stencil save/upgrade**: errors are logged internally or silently skipped.
- **HTMX template operations** (variant CRUD, version archive, deployment toggle, search): all get the same vague global banner regardless of what failed or why.

### Success feedback

- ~35 requests have **no success indicator** at all (silent swap/revert)
- ~40 rely on **HTMX DOM swap** being feedback; invisible if nothing visually changes
- 6 show **explicit feedback** (PDF preview opens, page title updates, alert popups)

### Suggested approach

1. Design a lightweight toast/notification component
2. Wire it into both the global HTMX error handler and a reusable `fetch()` wrapper
3. Retrofit all `fetch()` calls, starting with `settings.html` and `editor.html`
4. Replace `alert()` calls in `pdf-preview.js` and `version-comparison.html`

## Steps to Reproduce

Go to any template settings tab, toggle PDF/A on/off, change the theme, or rename the template. Throttle the network in DevTools: none produce user feedback on failure. Repeat across CRUD actions on any page -- the error behaviour varies.

## Expected Behavior

Every user-initiated async action should produce clear, consistent feedback: success on completion, a visible message on failure.

## Component

[...]

PR #830, opened 2026-08-15
refactor(ui): convert hand-rolled HTMX to native HTMX

whit3st:
## Description

Converts the template-settings controls and the version-comparison dialog trigger from hand-rolled HTMX to native HTMX.

The settings controls (default-theme select, template rename, PDF/A toggle) reimplemented HTMX by hand: `fetch()` PATCH calls with JSON bodies, a manually set `HX-Request` header, and manual DOM swaps of the returned fragments. Each control is now a plain `hx-patch`/`hx-trigger`/`hx-target`/`hx-swap` declaration backed by a focused form-encoded fragment endpoint (`PATCH …/name`, `…/theme`, `…/pdfa`) that re-renders from persisted state. The version-comparison dialog now loads via `hx-get` and the app-wide `data-open-dialog` hook instead of a per-page listener calling `htmx.ajax()` and `showModal()` by hand; running the comparison stays in JS because the previews are PDF blobs, which HTMX cannot swap.

Consequences and supporting changes:

- The rename response syncs the page header `<h1>` via an OOB fragment that targets a dedicated `#page-title-text` anchor, so the stage badge survives renames and `data-testid` stays a test-only contract. The OOB rides `innerHTML` (htmx has no `textContent` OOB style).
- The PDF/A endpoint reads the checkbox's literal value, not just param presence, and blank/unchanged renames stay a silent no-op re-render. A malformed `themeId` payload (not the select's `catalogKey/themeKey` shape) gets a 400.
- The JSON `PATCH /tenants/…/templates/{catalogId}/{id}` UI route and the orphaned `static/js/modules/api-client.js` are removed — no callers remain. UI-internal surface only; the REST API under `/api` is unaffected.
- JS keeps only keyboard UX (Enter commits, Escape reverts the rename input to its server-rendered `defaultValue`).
- Success paths are behavior-identical; one deliberate difference: a failed update no longer snaps the control back — the entered value stays until reload (previously a silent revert).
- Docs: the standalone-control convention is written down in `docs/htmx.md` (checkbox value semantics, server-side validation posture, OOB companions on dedicated ids, "native `hx-*` first; `fetch` only for blobs and editor callbacks"); `docs/pdfa.md` drops the section for the deleted JSON route and notes the setting is UI-only; `docs/refactor.md` marks the superseded api-client tasks.
- Tests: the settings PATCH endpoints' contract is pinned (fragment shapes, OOB title sync, 404 for unknown and cross-tenant templates, the non-HTMX redirect, PDF/A enable/disable transitions seeded from a known state), plus browser tests for the two rewired interactions (compare trigger's click firing both `hx-get` and dialog open; rename Escape-revert tracking the latest committed name).

## Related Issue(s)

Relates to #477

## Type of Change

- [ ] Documentation update
- [x] Refactoring (no functional changes)

## Component(s) Affected

- [x] Documentation
- [ ] CI/CD

## Checklist

## Additional Notes

The `hx-patch` control pattern established here (a lone control PATCHing a focused fragment endpoint that re-renders from persisted state) is documented in `docs/htmx.md` as the convention for future standalone controls.


apps/epistola/src/main/kotlin/app/epistola/suite/handlers/DocumentTemplateHandler.kt
```diff
@@ -495,59 +479,119 @@ class DocumentTemplateHandler(
...
-                pdfaEnabled = updateRequest.pdfaEnabled,
-            ).execute() ?: return ServerResponse.notFound().build()
+        return request.htmx {
+            fragment("templates/detail/settings", "output-settings") {
+                "tenantId" to tenantId.key
+                "catalogId" to catalogId.value
+                "template" to template
+                "editable" to (template.catalogType == app.epistola.suite.catalog.CatalogType.AUTHORED)
+            }
+            onNonHtmx {
```
whit3st: these should be `onFullPage`, the newer handler

docs/refactor.md
```diff
@@ -20,7 +20,7 @@ The refactoring focuses on:
```
whit3st: should this file tracked anymore?

docs/pdfa.md
```diff
@@ -34,14 +34,11 @@ This matters for regulated industries (finance, healthcare, government) and any
...
-PATCH /tenants/{tenantId}/templates/{templateId}
-Content-Type: application/json
-
-{ "pdfaEnabled": true }
-```
+There is no REST API endpoint for this setting — it is configured through the settings UI
+only. (The toggle itself issues a form-encoded `PATCH …/templates/{catalogId}/{templateId}/pdfa`
+UI request; UI routes are internal and may change freely.) The setting also does not
+round-trip through catalog export/import. Whether this UI-only scoping is deliberate is an
+open question tracked in [#817](https://github.com/epistola-app/epistola-suite/issues/817).
```
whit3st, line 48: This needs a verification before the PR merges as is. I would guess that this was a pure UI handler shaped to return JSON and its only consumer was frontend but better be certain, might break somewhere that I don't know.

apps/epistola/src/main/kotlin/app/epistola/suite/handlers/DocumentTemplateHandler.kt
whit3st: Fixed in 5a0d745fc — both new handlers (updatePdfa, updateName) now use onFullPage, and the test comment referencing the onNonHtmx branch is reworded to match. Pre-existing onNonHtmx usages in the file are left as-is (out of scope for this PR).

docs/refactor.md
whit3st: Agreed — deleted in 37842df4e. Audit showed every task is done or superseded: Task 1's split actually shipped in #156 (Variant/Version/TemplatePreview handlers exist in handlers/; what remains in DocumentTemplateHandler is the core CRUD the plan said to keep), Tasks 3/4/6/7 shipped despite still saying Pending, and 2/5 were already marked superseded. So no follow-up issue was needed — full task-by-task reasoning is in the commit body.

Issue #69, opened 2025-12-26
Add parameter usage tracking


sdegroot:
## Summary

Track which input parameters are used in template expressions through static analysis.

## Context

Understanding parameter usage helps with:
- Identifying unused parameters in schemas
- Detecting missing parameters in templates
- Providing usage analytics for schema design

## Acceptance Criteria

- [ ] Static analysis of template expressions
- [ ] Report of used parameters per template
- [ ] Identify unused schema parameters
- [ ] Detect expressions referencing undefined parameters
- [ ] API endpoint for usage report

## Technical Approach

Parse expression syntax `{{path.to.value}}` and correlate with dataModel schema.

PR #852, opened 2026-08-20
feat(editor): make list item spacing consistent and configurable

sdegroot:
## Dependency

Depends on merged [epistola-contract #69](https://github.com/epistola-app/epistola-contract/pull/69). This PR now consumes the published contract artifacts at version `1.1.0` in both the backend and editor.

## Summary

- add one inheritable `listItemSpacing` style across Text, Rich Text Block, and Data List
- default the gap between sibling list items to `0.5sp`, with `sp` and `pt` support
- use the same spacing behavior in the WYSIWYG editor and generated PDF
- make nested ProseMirror content respect the configured text line height
- introduce rendering defaults V2 while preserving V1 output for published templates
- demonstrate compact, default, and roomy list spacing in the Simple Letter catalog
- document the functional behavior in the changelog

## Contract 1.1.0 compatibility

- atomically update the backend, editor, and lockfile contract versions
- adapt to the new `CatalogInfo` construction API
- source current catalog wire-version constants from the contract
- update bundled catalogs, test fixtures, fingerprints, and assertions to schema 6

## Verification

- `./gradlew :modules:generation:ktlintFormat :modules:generation:test`
- editor Vitest suite against published `@epistola.app/epistola-catalog@1.1.0` — 2,051 tests passed
- editor typecheck, production build, and component-registry validation
- `./gradlew :modules:epistola-core:test testClasses checkContractVersionAlignment` — 1,270 core tests passed
- `git diff --check`


modules/editor/src/main/typescript/engine/registry.ts
```diff
@@ -1067,7 +1067,7 @@ export function createDefaultRegistry(): ComponentRegistry {
     category: 'logic',
     slots: [{ name: 'item-template' }],
     allowedChildren: { mode: 'all' },
-    applicableStyles: LAYOUT_STYLES,
+    applicableStyles: [...LAYOUT_STYLES, 'listItemSpacing'],
```
sdegroot: why is this expcliitly named here? its in the contract right?

sdegroot: Good catch. This metadata is owned by `@epistola.app/epistola-catalog`: `withContractMetadata()` overwrites the local `applicableStyles`, defaults, and examples at registration time. I removed the redundant local list-spacing metadata for Data List, Text, and Rich Text. The remaining suite changes are rendering logic and tests against the contract metadata. Editor tests, typecheck, build, and registry parity all pass.

